Issue #237, opened 2014-03-13
Prepare axis of rotation alignment procedure


tfarago:
Iterative using UFO FW.

PR #239, opened 2014-03-14
Initial implementation of iterative axis alignment

matze:
This PR addresses #237. Waiting for feedback from @tomago.

### DIFF
--- a/concert/ext/ufo.py
+++ b/concert/ext/ufo.py
@@ -8,7 +8,11 @@ try:
 except ImportError as e:
     print(str(e))
 
-from concert.coroutines.base import coroutine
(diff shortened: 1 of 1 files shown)


tfarago:
Still needs some work on the backprojection side, shifting.

tfarago:
:fireworks: now the algorithm actually works (the filtering needed to be changed to the padded one). I tested the `compute_rotation_axis`. I think it can be merged now.

matze:
Great :+1: 

But on the other hand, we could wait another two weeks and celebrate yet another birthday ;-)

tfarago:
This thread will be available also by then. Feel free to do so ;D

PR #258, opened 2014-05-09
graphical user interface

prybalko:
GUI for the concert. Lightsource widget works for dummy, the rest of widgets is templates. Further, I'm planning to create own classes for each widget. (It will look like LightSourceWidget class).

ConcertGUI class is a basic class in this application.
MyGroupBox is a class which responsible for behavior of device's widgets


matze:
@pavel-rybalko, @tfarago: Up to now we were pretty good at avoiding duplicating any work. With recent additions of visualization to the Concert and the Tofu GUI we are at a point where we should move data pre-processing (i.e. meshing etc.) and creating suitable PyQtGraph widgets to a single place. I don't think Concert is particularly suited for that, so I suggest we move that code to [tofu](https://github.com/ufo-kit/tofu).

If there are no objections, I will refactor the existing Tofu visualization code into `tofu.vis` and `tofu.vis.qt` respectively which you can then use as a template. Once done, we make Tofu a soft dependency here and will finally (re-) start the review and merge process.

tfarago:
Go for it. By the way there are even more possibilities to unduplicate code by using tofu in Concert (or separately in some concert session), e.g. correlation for rotation axis is implemented in both as well as flat correction, and who knows what comes next. That are all tools which fit much better someplace else than core Concert.

tfarago:
Todo:
* Widgets are able to reach "outside" of the canvas
* SlowLinearMotor `NotImplementedError`
* functions should indicate when they are done
* add expect decorator to every function from `processes`
* detailization -> level of detail

tfarago:
Can we merge this or should I also review?

matze:
You should at least test it once. In case you have you have my :+1:

MarcusZuber:
I would suggest to add the proper dependencies in requirements.txt.

tfarago:
@matze I would prefer not to have that in general requirements. Maybe we can document that or does something like optional requirements exist?

MarcusZuber:
I already found a stupid issue ;-)
When using a German locale where the decimal-seperator is a ',' the setting of a value fails because the value is not translated back properly. On the console I get: 
ValueError: invalid literal for float(): -56,26

prybalko:
It's sad to let it go

tfarago:
Yep. But you are very welcome to pick things up were you left them off after the Python 3 transition ;-)

PR #303, opened 2014-09-05
PyqtgraphViewer

prybalko:
Image viewer based on Pyqtgraph lib

### DIFF
--- a/concert/async.py
+++ b/concert/async.py
@@ -92,7 +92,7 @@ if concert.config.ENABLE_GEVENT:
         import gevent.monkey
         import gevent.threadpool
 
-        gevent.monkey.patch_all()
+        gevent.monkey.patch_all(thread=False, socket=False)
 
         # XXX: we have to import threading after patching
         import threading
--- a/concert/ext/viewers.py
+++ b/concert/ext/viewers.py
@@ -21,6 +21,7 @@ from concert.async import threaded
 from concert.quantities import q
 from concert.storage import write_tiff
 from concert.coroutines.base import coroutine
+import concert.config
 
 
 LOG = logging.getLogger(__name__)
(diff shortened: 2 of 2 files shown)


pavel-rybalko:
I've added a gevent support for the viewer, but it's kinda dirty and requires press "Enter" after entering a command

pavel-rybalko:
I've done some tests and measurements regarding performance of `Pyqtgraph` library.
I wrote a [script](https://gist.github.com/pavel-rybalko/27b0b5ceeac34cb682f8), which creates a viewer and updates it in another thread / process as fast as it can. The results of performance are presented in the following table (clickable).
selection_052
As you can see, the results are not really impressive and comparable with Matplatlib. 

However, I found another library for visualization , which is OpenGL-based library called [Vispy](https://github.com/vispy/vispy). And it works pretty fast ([here](https://gist.github.com/pavel-rybalko/3805d82630993d3c077c) is my script):
selection_051
But this library is still under development and has some crucial problems with memory leakage ([issue 1](https://github.com/vispy/vispy/issues/724), [issue 2](https://github.com/vispy/vispy/issues/735))

tfarago:
Nice summary, thanks. I would *not* go for Pyqtgraph and wait for vispy.

pavel-rybalko:
ToDo:
latency test vs Matplotlib

campagnola:
Sorry to interrupt; I got here by way of a related discussion on vispy. Your pyqtgraph test uses ImageView, which is not intended for video performance (the histogram computation and plotting is quite expensive). I recommend looking at `examples/VideoSpeedTest.py` for examples of the kind of framerates you can expect.

matze:
Thanks for the pointer. I hope @pavel-rybalko could redo the test before leaving back?

pavel-rybalko:
Here is an updated graph:
selection_012
I didn't write my own script for testing, just trustfully used `examples/VideoSpeedTest.py`

Note: I limited FPS axis an the graph by 100

campagnola:
That looks about right. Also note that the vispy framerates are probably limited by the GUI toolkit; their theoretical maxima should be much higher (so you will need much less CPU time to achieve the same framerate on vispy than other libraries).

pavel-rybalko:
I wrote a [script](https://gist.github.com/pavel-rybalko/27b0b5ceeac34cb682f8), which uses `GraphicsView`  and update images in another process. It works fine, but we need to integrate it into `RemoteGraphicsView` (which implements multithreading) or to implement it like in my script (explicit multiprocessing). But already I don't have time for that though.

PR #339, opened 2014-10-21
First attempt to use more gpus

tfarago:
A collaborative PR for making UFO work on more GPUs from concert.

### DIFF
--- a/concert/ext/ufo.py
+++ b/concert/ext/ufo.py
@@ -628,7 +628,7 @@ class UniversalBackprojectManager(object):
     def produce(self):
         sleep_time = self.projection_sleep_time.to(q.s).magnitude
         for i in range(self.args.number):
-            while self._in_index < i + 1:
+            while self._num_received_projections < i + 1:
                 time.sleep(sleep_time)
             yield self.projections[i]
 
@@ -675,7 +675,7 @@ class UniversalBackprojectManager(object):
         self._process_event.clear()
         LOG.debug('Backprojector manager start')
         st = time.time()
-        self._in_index = 0
+        self._num_received_projections = 0
         aborted = False
 
         def prepare_and_start():
@@ -684,7 +684,8 @@ class UniversalBackprojectManager(object):
                 LOG.debug('Waiting for events')
                 for event in wait_for_events:
                     event.wait()
-                LOG.debug('Waiting for events done (cached projections: %d)', self._in_index)
+                LOG.debug('Waiting for events done (cached projections: %d)',
+                          self._num_received_projections)
 
             self._update()
 
@@ -739,13 +740,13 @@ class UniversalBackprojectManager(object):
                     self.projections.dtype):
                 self.projections = np.empty(in_shape, dtype=projection.dtype)
             self.projections[0] = projection
-            self._in_index = 1
+            self._num_received_projections = 1
 
             while True:
                 projection = yield
-                self.projections[self._in_index] = projection
-                self._in_index += 1
-                if self._in_index == self.args.number:
+                self.projections[self._num_received_projections] = projection
+                self._num_received_projections += 1
+                if self._num_received_projections == self.args.number:
                     if wait_for_projections:
                         arg_thread = threading.Thread(target=prepare_and_start)
                         arg_thread.start()
@@ -753,11 +754,11 @@ class UniversalBackprojectManager(object):
                     if block:
                         self.join()
         except GeneratorExit:
-            if self._in_index < self.projections.shape[0]:
+            if self._num_received_projections < self.projections.shape[0]:
                 LOG.error('Backprojection manager has not obtained enough projections')
                 aborted = True
                 # Let UFO process fake projections until the graph can be aborted as well
-                self._in_index = self.projections.shape[0]
+                self._num_received_projections = self.projections.shape[0]
                 arg_thread.join()
 
 
--- a/concert/ext/ufo.py
+++ b/concert/ext/ufo.py
@@ -679,6 +679,7 @@ class UniversalBackprojectManager(object):
             if i == self.args.number:
                 LOG.debug('Last projection dispatched by manager')
                 if block:
+                    arg_thread.join()
                     self._pool.join()
 
 
--- a/concert/ext/ufo.py
+++ b/concert/ext/ufo.py
@@ -446,7 +446,8 @@ def compute_rotation_axis(sinogram, initial_step=None, max_iterations=14,
 
 
 class GeneralBackprojectArgs(object):
-    def __init__(self, width, height, center_position_x, center_position_z, number, overall_angle=np.pi):
+    def __init__(self, width, height, center_position_x, center_position_z, number,
+                 overall_angle=np.pi):
         self._slice_metric = None
         self._slice_metrics = ['min', 'max', 'sum', 'mean', 'var', 'std', 'skew',
                                'kurtosis', 'sag']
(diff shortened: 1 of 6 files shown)


tfarago:
I am trying to make this work with `CopyTask` but no luck. The code from [here](https://gist.github.com/a3a4dc1d647e409e26dc.git) crashes with the following awfully:
```python
f = FooProcess()
im = np.empty((512, 512), dtype=np.float32)
result = Result()
inject((im,), f(result()))
```
Can @matze take a look please?

tfarago:
Similar stuff written in "pure" Ufo works fine.

matze:
I will have a look.

tfarago:
I think I got it. I have to make as many output tasks as many I have "gpu branches", that seems to work, I will test it and if it *really* works you will see some more commits here.

matze:
Hmm, yes. Each pipeline (also part of a more complex graph) needs a "sink". I should really add a check so that these kind of graphs cannot be executed.

tfarago:
Now I added a more complex graph to the gist which segfaults. Could you have a look at the third revision please? We can continue the discussion there then.

tfarago:
Man, this took a while, but I finally see nice GPU utilization. Before I start, **this is a very rough prototype**. The reason I pushed it now is that I have to do other stuff but you could take a look at the issues I mention below in parallel. I will clean up the code as soon as I am free...

Preliminary results: For a dummy data set without reading and flat correction I get around 700 GUPS, i.e. 2016^3 volume from 3000 projections in 36 s. The online code is now actually 2x faster than the `tofu` version, see explanation below.
The key is to
- push alredy cropped projections to ufo, hence there is less data copying
- Set up separate reconstructor for every GPU and use separate thread for every reconstructor (this is not the case in `tofu` so far, hence the speedup)

The flat corrected version is slower, I think because I need to push dark and flat to ufo all the time. Can you have a look please @matze about what could be done? Also, after I call `wait()` on the `InjectProcess` the CPUs are working like crazy and I have no clue why. Can you check that as well please?

matze:
Here are some issues I noticed running it on compute2:

* I cannot exit the concert shell after creating a `UniRecoManager` object without resorting to <kbd>Ctrl</kbd><kbd>C</kbd> or killing the process. However, exiting *after* reconstruction works fine :confused: 
* After reconstruction GPU memory is still allocated, a second run is thus impossible.
* It's pretty fast and takes ~ 64 seconds on that rather ancient machine (2x K20Xm, 4x GTX TITAN). In fact, if I just use the TITANs it's just 40s and quicker than the "standard" [reconstruction](https://tomopedia.github.io/software/ufo#benchmarks) (fifth line in the table) :astonished: . Stupid question: is really the entire volume computed?
* On this machine the CPUs are not going crazy after `wait()` …

tfarago:
> I cannot exit the concert shell after creating a UniRecoManager object without resorting to CtrlC or killing the process. However, exiting after reconstruction works fine

It's because the reconstructors are created and *started* when you create the manager. I will defer the starting to the real starting of the coroutine.

> After reconstruction GPU memory is still allocated, a second run is thus impossible.

I just found out that when you delete the reconstructors with `del manager.reconstructors` the memory is in fact deallocated. Although it seems there are a couple of hundred MB still leaking. However, I cannot test that now because ufo2 was shifted to our server room and the network is not yet working. I will get back to this point.

> It's pretty fast

Well that was my aim. I checked with real data and got the full volume correctly, as with the offline version. The separate threads thing in concert + the fact that we can use the full GPU memory reduces the copy/compute overhead a *lot*. Check nvidia-smi while it's reconstructing, you should see a quite steady utilization of more than 90 %.

> On this machine the CPUs are not going crazy after

on ufo2 and ankatoto-concert they are... Almost 100 % CPU utilization even after wait and *deletion* of the reconstructors and the manager...

tfarago:
I deferred the graph creation for when the coroutine starts, but it takes 20 on ufo2 just to create all the reconstructors, so it might be a good idea to really create them in the constructor because the way it is now will stall execution. Is this an OpenCL thing or is it on our side?
I also delete the reconstructors now because of the leak, but even when I do this there is 500 MB leaking on ufo2.

tfarago:
Could you first try to change the `InjectProcess` in such a way that I don't need to insert dark and flat all the time? It really hurts performance...

tfarago:
Ha! That was easy :smile:

tfarago:
Man, the `projection.astype(np.float32)` is also slow.

concert/ext/ufo.py
```diff
@@ -152,7 +152,8 @@ def insert(self, array, node=None, index=0):
         else:
             self.ufo_buffers[node][index] = self.input_tasks[node][index].get_input_buffer()
 
-        self.ufo_buffers[node][index].copy_host_array(array.__array_interface__['data'][0])
+        if array is not None:
+            self.ufo_buffers[node][index].copy_host_array(array.__array_interface__['data'][0])
```
tfarago: I tried
```python
if array.dtype != np.float32:
    self.ufo_buffers[node][index].copy_host_array(2)
```

in order to delegate the float conversion to ufo but when I use more GPUs I get a segfault. Moreover, I cannot use the enumeration, because in python the prefix is stripped off and `Ufo.BufferDepth.16U` is invalid. I can work around this with `getattr(Ufo.BufferDepth, '16U')` but that's ripe for a facepalm... :smile:

tfarago: Actually, like we have `ufo_buffer_convert` now, we could in principle delegate this to the filters and write a GPU task which does that. The host buffer is always allocated for float in the beginning, right? In that case when I ask for a host buffer in my task an treat the array appropriately in OpenCL it should be possible. Or am I missing something except it's magicky?

tfarago:
OK so right now it's quite OK from the speed point of view (48 s for 2016 cube from 3000 projections *with flat correction*). Unfortunately there is the memory leak problem. What is even worse is that even when I have enough memory and execute the graphs (create new managers) more times I get strange OpenCL errors.

concert/ext/ufo.py
matze: There is such a [task](https://github.com/ufo-kit/ufo-filters/blob/master/src/ufo-memory-in-task.c) but as of now it's doing conversion on the CPU (no idea, if it would actually be faster to do it on the GPU, I doubt it). And yes, the host buffer allocates float arrays *if* required.

tfarago:
81c71b16efd2f260be9c388f1d536f727f362eac kills two birds with one stone

- there is no memory leak anymore
- since I reuse the resources, the OpenCL initialization takes place only once upon manager creation, so it doesn't matter for running the reconstructors

tfarago:
I will start with cleanup and creating and Addon for online reconstruction, next week we should be able to test this at the beam line.

concert/ext/ufo.py
```diff
@@ -452,58 +454,53 @@ def __init__(self, width, height, center_x, center_z, number, overall_angle=np.p
...
-class UniReco(InjectProcess):
-    def __init__(self, args, flat=None, dark=None, gpu_index=None, x_region=None, y_region=None,
-                 region=None, resources=None):
+class UniversalBackproject(InjectProcess):
+    def __init__(self, args, resources=None, gpu_index=None, flat=None, dark=None, region=None):
         scheduler = Ufo.FixedScheduler()
         if resources:
             scheduler.set_resources(resources)
         gpus = scheduler.get_resources().get_gpu_nodes()
+        gpu_index = 0 if gpu_index is None else gpu_index
```
tfarago: Good that I named the commit *start* of the cleanup... 😄

tfarago:
@matze even though there is no memory leak anymore I still see some threads hanging around saturating the CPUs (both on ufo2 and ankatoto-concert machine). There are as many threads as there are backprojectors in the manager. I checked and the python threads all exit as they should once the processing is finished. Even `threading.enumerate()` doen't show anything suspicious, that's why I think the problem is on the UFO, or worse, on the driver side. Can you check this please?

tfarago:
This has been around for long enough.

PR #351, opened 2015-01-23
Added optinal parameters to devices.cameras.uca interface

TimoDritschler:
Changed the ```uca.__init__``` to accept optional parameters and use the new ```get_camerah``` interface to create the camera instance. When no parameters are supplied, this new interface behaves just like the old ```get_camerav```


matze:
Looks good :+1:

PR #354, opened 2015-01-30
Fix #353: Prevent changing parameters while recording

### LINKED ISSUE #353: Crash when calling grab() on an uca camera after increasing its ROI
Example:
```python
from concert.devices.cameras import uca
cam = uca.Camera('mock')
cam.start_recording()
cam.roi_width = 1024 * q.pixel
cam.grab()
Segmentation fault
```
This is probably caused by the internal _numpy array_ used to pass to the ```_grab_real```. The array is being constructed with the size information from ```self._record_shape```  (see [here](https://github.com/ufo-kit/concert/blob/517cefa187e05b71133a956e0258ce8cc351d185/concert/devices/cameras/uca.py#L186)) which is only updated by ```_record_real```. Since the size information does not get updated when changing the ROI directly, the array will be too small for the expected frame and crash.

TimoDritschler:
Should 'fix' #353 by not allowing to change any parameter while recording, in the first place.

### DIFF
--- a/concert/devices/cameras/uca.py
+++ b/concert/devices/cameras/uca.py
@@ -15,10 +15,9 @@ LOG = logging.getLogger(__name__)
 
 def _new_setter_wrapper(name, unit=None):
     def _wrapper(instance, value):
-        if instance:
-            if instance.state == 'recording':
-                raise base.CameraError(
-                'Changing parameters is not allowed while recording')
+        if instance.state == 'recording':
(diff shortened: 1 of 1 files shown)


TimoDritschler:
I don't know. I was just being careful.

matze:
The remaining code wouldn't have worked if that was the case. So I suggest you remove that check and then you can merge.

PR #356, opened 2015-02-04
[WIP] Print exceptions of async calls

MarcusZuber:
A callback will print exceptions of async functions.
This can be enabled and disabled via concert.config.ENABLE_PRINT_ASYNC_EXCEPTION.

### DIFF
--- a/concert/async.py
+++ b/concert/async.py
@@ -215,7 +215,8 @@ if not concert.config.ENABLE_GEVENT or not HAVE_GEVENT:
                 @functools.wraps(func)
                 def _inner(*args, **kwargs):
                     future = EXECUTOR.submit(func, *args, **kwargs)
-                    future.add_done_callback(print_exception)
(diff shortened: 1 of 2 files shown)


MarcusZuber:
Yes. I realized it a few seconds after the commit. I have to work on somethin else now, but I will proceed with it in the near future.

matze:
If you could come up with a tiny test, I'd say it's ready for a merge. What about you @tfarago?

MarcusZuber:
I want to include the implementation for gevents as well. For now it works only with concurrent.futures.

tfarago:
Let's wait for the gevent, the tests and maybe a line in the docs and we're good to go.

MarcusZuber:
Beside that a new input line after the printing would be nice it works for futures and gevent-futures.
But I have no idea to write a test to check if the printing happend.

tfarago:
I think after asyncio this will be obsolete, if not, re-open.

PR #360, opened 2015-02-09
Don't use upper bound on `measure` test duration

tfarago:
Sometimes it is more than those 10 milliseconds.


tfarago:
@matze OK?

matze:
OK! :+1:

PR #362, opened 2015-02-13
Check emergency stop

tfarago:
This is going to be useful when an actual damage can be done and there are mechanisms preventing that. If those mechanisms are exposed to Concert we can react accordingly.

### DIFF
--- a/concert/session/utils.py
+++ b/concert/session/utils.py
@@ -143,7 +143,7 @@ def code_of(func):
 
 
 def abort():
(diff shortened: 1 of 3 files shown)


tfarago:
If @matze doesn't know any better way than `os.abort()` then I can S & M. BTW your remarks are there already.

MarcusZuber:
:thumbsup:
I tested it in the lab and it seems to work as expected (and requested by Thomas).

matze:
No, I don't know any better :+1:

PR #364, opened 2015-02-18
[WIP] Fix #272: emit dispatcher signal after parameter setting

### LINKED ISSUE #272: Shall we use the dispatcher again?
For @pavel-rybalko gui it would of utmost convenience. On the other hand there might be a reason why we removed it from params and all the other parts of concert. @matze what is your opinion on this? I don't see any other way how he could easily listen to devices and make changes to the gui in an automatic way.

prybalko:
Emit the dispatcher signal "value_changed" every time device parameter is changed

### DIFF
--- a/concert/base.py
+++ b/concert/base.py
@@ -9,6 +9,7 @@ import threading
 from concert.helpers import memoize
 from concert.async import async, wait, busy_wait
 from concert.quantities import q
+from concert.async import dispatcher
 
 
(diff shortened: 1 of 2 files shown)


matze:
Just a general question: does this change really affect #272?

pavel-rybalko:
As I understand, we need either to get back to dispatcher, make something instead of it or renounce the use of it. That is not an answer to the question, but more like a step towards the first option.

PR #369, opened 2015-05-27
Add Flask-like extension system

matze:
This allows third-party extensions to be imported from a common namespace. By
ensuring use of setuptools we can also enforce proper versioning and deployment
strategies.


matze:
@tfarago: right now the umbrella namespace is `concert.third` because `ext` is already used by stuff that we ship. If you have a better name … tell me :smile:

tfarago:
a) :+1: and b) no I don't...

PR #370, opened 2015-06-09
Generalize acquisitions

tfarago:
It makes sense to split data acquisition and download. Especially when one wants to acquire all the data at once (e.g. fill up a camera buffer with different frame types by external trigger) and worry about the download later.

### DIFF
--- a/concert/experiments/base.py
+++ b/concert/experiments/base.py
@@ -16,11 +16,11 @@ class Acquisition(object):
     """
     An acquisition acquires data, gets it and sends it to consumers.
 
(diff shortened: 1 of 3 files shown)


tfarago:
Rebased on the fixed master, let's hope the tests pass.

matze:
If you could add an example to the docs so that I can hook that up myself in the future without twisting my brains, I'd give you a :+1:

tfarago:
I added something to the docs, hope it's fine.

matze:
Thanks! :+1:

PR #372, opened 2015-06-12
Create session manual from docstring

matze:
This change adds a `docs` command to create a PDF manual from a session's docstring written in Markdown. That means

```
"""# *foo* session manual

## Usage

Start an experiment acquisition with

    ex.acquire('foo').join()

where `foo` denotes the data set name.

## Notes 

This *is* **ordinary** [Markdown](http://disney.com)
"""
```

turns into

concertdocs

### DIFF
--- a/bin/concert
+++ b/bin/concert
@@ -450,6 +450,35 @@ class StartCommand(Command):
             print(msg.format(exception))
 
 
+class DocsCommand(Command):
+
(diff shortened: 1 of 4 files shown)


tfarago:
Very nice :+1:

tfarago:
@MarcusZuber this might be interesting for you too.

PR #374, opened 2015-09-15
Implementation of velocity calculation for continuous motors

MarcusZuber:
The _get_velocity command now calculates the velocity by measuring the
motor position twice and calculates the current velocity out of these
values since many motors do not feature a measured velocity.

Open for discussion of a better solution!

### DIFF
--- a/concert/config.py
+++ b/concert/config.py
@@ -9,13 +9,22 @@
     Turn on gevent support. If geven is not available, fall back to
     ThreadPoolExecutor approach.
 
+.. data:: MOTOR_VELOCITY_SAMPLING_TIME
+
+    Time step for calculation of motor velocity by measuring two postion
+    values. Longer values will create more acurate results but reading the
+    velocity will take more time.
+
 .. data:: PROGRESS_BAR
 
     Turn on progress bar by long-lasting operations if tqdm package is present
 """
+from concert.quantities import q
 
 ENABLE_ASYNC = True
 ENABLE_GEVENT = False
 # Prints the exception source by fake futures
 PRINT_NOASYNC_EXCEPTION = True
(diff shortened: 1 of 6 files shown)


tfarago:
I like it. We should reflect the changes in the dummy motors, however. They could use the `time` module to compute their position based on velocity and then the `_get_velocity` implementation can be dropped completely from the dummy motors. I will write the tests when that's done.

MarcusZuber:
It seams to work. I'm open for comments!
Should the LinearMotor and RotationMotor also get a readable (calculated) velocity or only the Continuous ones?

tfarago:
I think we should keep the simple motors really simple if there is no real need to stuff them with functionality, so no. They can even return 0 as far as I am concerned.

MarcusZuber:
Just found out, that one thing is still missing. The velocity can not be set in the dummy motors. I will include it in the next days.

MarcusZuber:
@tfarago If you could help me setting up the test properly we could finish this.

tfarago:
I would first deal with #435, then adjust code here accordingly and proceed. In general it's fine.

tfarago:
@MarcusZuber have you reviewed this since #435 was merged?

MarcusZuber:
Not yet. But basically one must change the limits from +- inf to None. The rest should be fine.

MarcusZuber:
Ok, it seams to be a bit broken. I will rebase it and fix it.

MarcusZuber:
OK, now it should work out. The only failing tests are already in the master branch :flushed: failing.

tfarago:
OK, the failing tests have nothing to do with this PR, merging.

PR #376, opened 2015-11-16
Numpy accumulate

tfarago:
It will be very convenient to have the data accumulated already as a numpy ND stack when implementing general backprojectors or data processing routines in general. With this we can do for example `backproject.insert(accumulate.items[:, height, :])` which wouldn't be so simple AFAIK with a list of 2D numpy arrays.

### DIFF
--- a/concert/coroutines/sinks.py
+++ b/concert/coroutines/sinks.py
@@ -34,21 +34,41 @@ def null():
 
 
 class Accumulate(object):
-    """Accumulate items in a list."""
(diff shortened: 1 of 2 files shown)


tfarago:
@matze if there isn't a more elegant way this is finished.

matze:
No, looks like  a good idea. However, you should also check the values in the test not only the type. Better safe than sorry.

matze:
From my point of view, we can merge this :+1:

PR #377, opened 2015-11-16
Accumulate addon

tfarago:
Stores experiment data so one can easily and efficiently access all of it as numpy arrays.

Todo:
* [x] make the interface more compact
* [x] make sure memory is released on `detach`

Requires #376. With this addon we can do:
```python
ex = Experiment(...)
acc = Accumulator(ex.acquisitions)
ex.run().join()
backproject.insert(acc.items[ex.radios][:, height, :])
```


tfarago:
@matze this one is done too.

matze:
:+1:

PR #378, opened 2015-11-17
Print noasync traceback

### LINKED ISSUE #301: Disabling async is broken
Running `runtests.py --disable-async` breaks many tests when calling the wrapped function.

tfarago:
Fixes #301 again and gives the possibility to suppress the exception source.


matze:
Looks good! :+1:

PR #379, opened 2015-11-18
Fix FlatCorrectedBackproject

tfarago:
Now the backprojector works, I will make a simple session now to check consistency of results.

### DIFF
--- a/concert/ext/ufo.py
+++ b/concert/ext/ufo.py
@@ -128,8 +128,7 @@ class InjectProcess(object):
                 node = self.input_tasks.keys()[0]
         if self.ufo_buffers[node][index] is None:
             # reverse shape from rows, cols to x, y
-            shape = array.shape[1], array.shape[0]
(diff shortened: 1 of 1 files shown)


tfarago:
Deals with #380

matze:
If the test passes – which I don't doubt – you can merge :+1:

tfarago:
Wait wait, let's wait for the other part which is inconsistency.

tfarago:
On the camlin machine you have a `check-fcbp` session and you can do both runs fc and nofc and see the results are wrong. I guess there is something wrong with the 1D array passing to the framework, can you check that please?

matze:
The flat correction works as expected, i.e. the resulting sinogram has much less vertical artifacts and appears more even. Trying now to guess what happens passing on to the reconstruction.

tfarago:
Hmm, that would have been my guess where the problems come, but OK. Let's see what happens down the road.

tfarago:
The obvious problem is fixed here, inconsistencies in reconstruction will be dealt with elsewhere.

PR #381, opened 2015-11-19
Fix potential segfault

matze:
Passing the NumPy data pointer to the buffer does not work at least on one machine. Since this always seemed to be to brittle, we added a new call to ufo-core which allows data to be written to the host array. This means an additional data copy but this is effectively the same as in the very first version of the InjectProcess.

Note, though that this requires current HEAD (i.e. v0.10) of ufo-core.


tfarago:
I can live with that, we can optimize later.

PR #383, opened 2015-11-20
Add wait time to Dimax's start_recording

### LINKED ISSUE #382: Dimax start_recording stalls
When I use slow frame rate then after I call `dimax.start_recording()` it takes some time until the camera *actually* starts recording. The problem is, the uca camera state is only in stored in software so I don't have a real feedback from the camera (I hope I remember it right). Consequently, when I ask for a frame I get a timeout because the camera is not ready, e.g. when using [frames](https://github.com/ufo-kit/concert/blob/master/concert/experiments/imaging.py#L6).
I can imagine that libuca calls the appropriate function on the camera but doesn't ask for an ACK. @matze is it possible to get such an ACK? If not I don't see an elegant way to solve this, especially when the actual delay depends on the frame rate.

tfarago:
I know, but what else should we do?

fixes #382

### DIFF
--- a/concert/devices/cameras/pco.py
+++ b/concert/devices/cameras/pco.py
@@ -5,6 +5,7 @@ from datetime import datetime
 import numpy as np
 from concert.devices.cameras import base
 from concert.devices.cameras.uca import Camera
+from concert.quantities import q
 
 
 class Pco(Camera):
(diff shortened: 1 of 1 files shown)


matze:
> I know, but what else should we do?

Get another job :smile: :+1:

PR #387, opened 2016-04-11
Extend queue

tfarago:
With this I get a bump in download bandwidth for the dimax from ~140 MB/s to ~210 MB/s.

### DIFF
--- a/concert/coroutines/filters.py
+++ b/concert/coroutines/filters.py
@@ -6,6 +6,7 @@ except ImportError:
 import logging
 import time
 import numpy as np
+import concert.config as cfg
 from concert.quantities import q
 from concert.imageprocessing import ramp_filter
 from concert.async import threaded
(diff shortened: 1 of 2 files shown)


tfarago:
@matze just attracting your attention.

tfarago:
BTW what's up with Travis?

matze:
That's why I was reluctant to do anything, thought you would care about that issue with Gevent.

tfarago:
I lost the failure between all those warnings. I am on it now. We should fix those too.

tfarago:
Which version of gevent do I actually need to make your `Greenlet` subclass work? I just installed the 1.1.1 version (just what pip gave me) and I get an error when the subclass tries to assign [kwargs](https://github.com/ufo-kit/concert/blob/master/concert/async.py#L119).

concert/coroutines/filters.py
```diff
@@ -61,6 +61,12 @@ def queue(consumer, process_all=True, block=False):
     all items in the serve loop are processed, *process_all* must be True as well for this to take
     effect.
     """
+    from concert.async import HAVE_GEVENT
+    if cfg.ENABLE_GEVENT and HAVE_GEVENT:
+        from gevent.event import Event
+    else:
+        from threading import Event
+
```
tfarago: Is `Event` seriously not patched??

tfarago:
OK, should be fixed by the last ugly commit.

matze:
Urgs, but okay :+1:

PR #389, opened 2016-04-13
Add flat field correction task

tfarago:
I wonder how did we not need this before...

### DIFF
--- a/concert/ext/ufo.py
+++ b/concert/ext/ufo.py
@@ -292,6 +292,39 @@ class FlatCorrectedBackproject(Backproject):
         consumer.send(self.result())
 
 
+class FlatCorrect(InjectProcess):
+
+    """
+    Coroutine to flat-field correct projections.
+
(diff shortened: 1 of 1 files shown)


concert/ext/ufo.py
```diff
@@ -292,6 +292,39 @@ def _process(self, sinogram, consumer):
...
+    @coroutine
+    def __call__(self, consumer):
+        """Get a sinogram, do filtered backprojection and send it to *consumer*."""
+        if not self._started:
+            self.start()
+
+        while True:
+            image = yield
+            self.insert(image.astype(np.float32), node=self.flat_correction, index=0)
+            self.insert(self.dark, node=self.flat_correction, index=1)
```
tfarago: This is very inefficient and help from @matze would be appreciated. I tried not to inserting all the time but then it was hanging.

tfarago: Have you seen this already, @matze ?

PR #390, opened 2016-04-21
Make parameter log output useful

tfarago:
Currently, the parameter setting log output is useless. It only says which class changes a parameter. If we have e.g. many `LinearMotor` instances the output looks like this:

``` bash
concert.base: foo: set LinearMotor::position='12 millimeter'
```

This PR changes that to the following:

``` bash
concert.base: foo: set LinearMotor:name::position='12 millimeter'
```

where _name_ is the instance name, you would get it after typing `name = LinearMotor()` in your session.

The implementation is quite nasty but with a very elegant outcome, no additional attributes like `name` have to be added to devices, reducing the work of device developers and users to 0. I think it's worth it. What is @matze 's opinion?

@MarcusZuber could you test it at the lab please to see if it works as expected?

### DIFF
--- a/bin/concert
+++ b/bin/concert
@@ -411,6 +411,7 @@ class StartCommand(Command):
             self.run_shell(module)
 
     def run_shell(self, module=None):
+        import concert
         from concert.base import (UnitError, LimitError, ParameterError,
                                   ReadAccessError, WriteAccessError, LockError)
 
@@ -435,6 +436,7 @@ class StartCommand(Command):
 
             if not module:
                 from concert.quantities import q
+            concert.base._get_instance_variable_name.memo = {}
 
             try:
                 shell = get_ipython_shell(config=config)
--- a/concert/base.py
+++ b/concert/base.py
@@ -312,9 +312,10 @@ class Parameter(object):
                 except AccessorNotImplementedError:
                     raise WriteAccessError(self.name)
 
-            msg = "set {}::{}='{}'"
-            name = instance.__class__.__name__
-            LOG.info(msg.format(name, self.name, value))
+            msg = "set {}:{}::{}='{}'"
+            cls = instance.__class__.__name__
+            name = _get_instance_variable_name(instance)
+            LOG.info(msg.format(cls, name, self.name, value))
         except KeyboardInterrupt:
             cancel_name = '_cancel_' + self.name
             if hasattr(instance, cancel_name):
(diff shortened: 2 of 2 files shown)


bin/concert
```diff
@@ -435,6 +436,7 @@ class StartCommand(Command):
 
             if not module:
                 from concert.quantities import q
+            concert.base._INSTANCE_CACHE = True
```
tfarago: I want to ensure nothing is cached before the devices are fully created and a session is started, because there might be some device initialized inside concert or elsewhere with the user-facing name defined in a session later. E.g. the dummy camera sets its attributes in a constructor, which would result in a name *self* which is not what we want. On the other hand, we need caching because the search is pretty slow, in the order of 10 ms.
The question is if this is the correct place to indicate a session has started.

concert/base.py
```diff
@@ -860,3 +861,61 @@ def unlock(self):
         """Unlock all the parameters for writing."""
         for param in self:
             param.unlock()
+
+
+_INSTANCE_CACHE = False
+
+
+def _memoize_instance_id(func):
```
matze: Just wondering, but is there a very urgent reason to implement that behaviour as a decorator?

matze:
Well the implementation is indeed ... interesting, but I suppose there's no other way anyway. So :+1:

concert/base.py
tfarago: Of course not, I was so happy I found the `memoize` in helpers that I wanted to use it but only later I realized I need a new one but to rethink the decorator usage somehow wasn't scheduled in my head :smile: 
Anyway, now it's there decoratorless.

tfarago:
If it's alright I can S & M.

tfarago:
What is wrong with Travis again? This has nothing to do with walkers...

matze:
If you found the problem :+1:

tfarago:
I will wait for Marcus and merge if it works for him.

MarcusZuber:
I my installation it is not working  :disappointed:.
`_get_instance_variable_name` returns always an empty string:

```python
l = LinearMotor()
_get_instance_variable_name(l.position)
''
```

tfarago:
Do `l.position = 1 * q.mm` and take a look in the logs. What does that say?

MarcusZuber:
This is the log output from the currently running experiment:
`[2016-04-25 16:54:45,438] INFO: concert.base: all-axes: set ContinuousRotationMotor:::position='149.23828125 degree'`

tfarago:
That's a wrong log, you use `LinearMotor` in your example, try to find that one please.

MarcusZuber:
Strange, now it works:
`[2016-04-25 16:57:32,658] INFO: concert.base: all-axes: set LinearMotor:l::position='1 millimeter'`

So we have to figure out why it does not work with the implementation of my motors...
I create the instance of my motor in an additional file which I load with `from motors import *` in my session file. If I understand your "magic" correctly the motors should still be findable in the stack.

tfarago:
For me this setting works as well. Give me a call when you are in the lab please and we can have a look together.

tfarago:
This has been implemented elsewhere.

PR #391, opened 2016-05-30
Forbid addon reattachment

tfarago:
If attached more times, addons cause trouble. This prevents that without the `Addon` developers needing to care about that.


tfarago:
@MarcusZuber once this is merger please check the test session again, for me it works now.

tfarago:
@matze is this OK?

matze:
Yeah, go ahead :+1:

MarcusZuber:
Oh, I was to slow... But its working fine!

PR #395, opened 2016-09-28
Fix #385: Optional output of resolve without units

### LINKED ISSUE #385: Output of resolve (not urgent)
Hi everybody,

as discussed with Tomas today, it would be great to have an option to get rid of unit of measurements and obtain values directly instead of an objects.

Cheers,
Sergei.

MariaMatveeva:


matze:
@milkdrop: if this is still an issue, please open #385 again.

PR #401, opened 2017-01-24
Fix #400: Fix properties mixup of uca-cameras

### LINKED ISSUE #400: Mixup of properties when multiple instances of uca-cameras (different uca plugins) are created
When instances of different uca-cameras are created it seems, that the newer camera also has properties of the older instances.

`from concert.devices.cameras.uca import Camera`
`cam_a = Camera('mock')`
`cam_b = Camera('file')`
`cam_c = Camera('mock')`

`cam_a` seems to work as expected, `cam_b` has an unimplemented property `degree-value` and `cam_c` has the unimplemented `path` property.

`cam_X.uca.props` shows only the correct properties of each device.

MariaMatveeva:


matze:
Very nice!

Issue #402, opened 2017-04-27
PyplotImageViewer delays Experiment execution


MarcusZuber:
When a PyPlotImageViewer is attached to an experiment the frame-rate drops. As far as I understood, the PyplotImageViewer should drop frames when the drawing of an old frame is not yet finished.

Example:
```python
from concert.experiments.base import Acquisition, Experiment
from concert.experiments.addons import Consumer
from concert.devices.cameras.dummy import Camera
from concert.ext.viewers import PyplotImageViewer
from concert.quantities import q

class MyExp(Experiment):
   def __init__(self):
      from concert.experiments.base import Acquisition, Experiment
      acq = Acquisition("img", self.take_frames)
      Experiment.__init__(self, [acq])

   def take_frames(self):
      import numpy as np
      from time import time
      start = time()
      for i in range(100):
         yield np.zeros((2048, 2048))
      print time()-start

exp = MyExp()
viewer = PyplotImageViewer()
live_view = Consumer(exp.acquisitions,viewer)
```

Running `exp.run()` like this gives me 1.24s. If I detatch the `live_view` I get 0.0125s.
I tested it on my local machine (no ssh x-forwarding).

Can I somehow make the PyplotImageViewer less blocking? Also the asyncronous ImageWriter delays the acquisition (until now usually less than a frame with 15fps and 2k x 2k pixels). Is it possible to speed up the whole addon-consumer-pipeline?

@tfarago I thinkt this is also related to the delay that we had (have) with the pco.4000

tfarago:
I have very inconsistent results with `np.zeros`. To me it seems that sometimes python optimizes that just for array *reservation* and sometimes it really allocates zeros. Try the following:
```python
class MyExp(Experiment):
   def __init__(self, num_iterations=30, create_new=True):
      acq = Acquisition("img", self.take_frames)
      Experiment.__init__(self, [acq])
      self.num_iterations = num_iterations
      self.create_new = create_new

   def take_frames(self):
      start = time()
      create_duration = 0
      im = np.zeros((2048, 2048))

      for i in range(self.num_iterations):
        create_st = time()
        if self.create_new:
            im = np.random.normal(size=(2048, 2048))
        create_duration += time() - create_st
        yield im

      print 'Creation duration:', create_duration
      print ' Overall duration:', time() - start

```

PR #403, opened 2017-04-28
Speedup imageviewer

tfarago:
An attempt to speed up both the image viewer and the main concert thread by

- enabling downsampling
- enabling specification of sleep_time=None

Deals with #402 

@MarcusZuber can you test this please?

### DIFF
--- a/concert/ext/viewers.py
+++ b/concert/ext/viewers.py
@@ -93,6 +93,7 @@ class PyplotViewerBase(object):
         self._updater = None
         self.view_function = view_function
         self._blit = blit
+        self._num_drawn = 0
 
     def __call__(self, size=None):
         """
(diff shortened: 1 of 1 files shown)


tfarago:
@MarcusZuber is there something against this being merged?

MarcusZuber:
Did you test it recently?

I think 1ms default sleep time  is a little bit optimistic. Increasing it should also help to improve the performance (maybe 16ms, then it is still ~60fps).

Didn't we exactly try this a few weeks ago (without using this implementation)?

tfarago:
> Did you test it recently?
Of course not.
> 
> I think 1ms default sleep time is a little bit optimistic. Increasing it should also help to improve the performance (maybe 16ms, then it is still ~60fps).
I don't think it will change a lot but I will test different numbers.
> 
> Didn't we exactly try this a few weeks ago (without using this implementation)?
No, that was skipping frames, that's something else.

PR #404, opened 2017-09-12
coroutines: Fix queue's occasional hanging

tfarago:
If the generator exited after the last item in the queue had been processed it
might have blocked forever waiting for a new item to come.

### DIFF
--- a/concert/coroutines/filters.py
+++ b/concert/coroutines/filters.py
@@ -51,7 +51,7 @@ def downsize(consumer, x_slice=None, y_slice=None, z_slice=None):
 
 
 @coroutine
-def queue(consumer, process_all=True, block=False):
+def queue(consumer, process_all=True, block=False, timeout=0.1*q.s):
     """
     queue(consumer, process_all=True, block=False)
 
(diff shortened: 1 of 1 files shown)


matze:
I merged a quick fix but this PR should still merge cleanly.

tfarago:
@MarcusZuber this could be interesting for you.

MarcusZuber:
I'm looking forward to it. But I still don't get why the old implementation did not work in all cases.

tfarago:
I boiled it down to this:
```python
def make_images():
    for im in images:
        yield im
    time.sleep(1)


def write_async(process_all=True, block=True):
    coro = walker.write()
    coro = queue(coro, process_all=process_all, block=block)

    inject(make_images(), coro)
```

For some reason, the `GeneratorExit` exception is emitted only *after* the `time.sleep(1)`, i.e. that the consumer thread in the `queue` coroutine gets ahead and waits for another item by the `queue.get()` which never comes.

tfarago:
It actually makes sense, because the whole *function* is a generator, hence, it exists when the function exits and not just the for loop.

tfarago:
This has been solved elsewhere 1b471108aac37051f16b0f939cf6b78a8e4338c2

PR #408, opened 2018-02-01
Enable multi-page image writing

tfarago:
It's time, the online reconstruction chokes the disks pretty bad, so this is actually essential to have.

### DIFF
--- a/concert/storage.py
+++ b/concert/storage.py
@@ -4,6 +4,7 @@ import logging
 import tifffile
 from logging import FileHandler, Formatter
 from concert.coroutines.base import coroutine, inject
+from concert.writers import TiffWriter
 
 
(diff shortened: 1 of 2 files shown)


tfarago:
I always think in terms of how much trouble will I have opening it afterwards and that depends on size, not number of images. I don't know, I just find it more natural to talk about bytes when it comes to files than to talk about numbers of images. Plus if the user specifies number of images, I would need to recompute the number of bytes based on the first image size and the number of images in the coroutine, of course it's no trouble, just a matter of taste.

tfarago:
So what with this? I would like to have it in master. Do you still want to specify images instead of bytes? If so, I will quickly do it and would merge...

matze:
No no, just merge it. You gave me your opinion :-)

PR #409, opened 2018-04-04
Uca camera state read from camera

MarcusZuber:
This reads the state of the camera from the uca device instead storing it in the local concert session.
This helps when using uca-net devices where the state can be out of sync to the state concert assumes.

### DIFF
--- a/concert/devices/cameras/uca.py
+++ b/concert/devices/cameras/uca.py
@@ -191,10 +191,7 @@ class Camera(base.Camera):
 
     @_translate_gerror
     async def _record_real(self):
-        self._record_shape = ((await self.get_roi_height()).magnitude,
-                              (await self.get_roi_width()).magnitude)
-        self._record_dtype = (np.uint16 if (await self.get_sensor_bitdepth()).magnitude > 8 else
-                              np.uint8)
+        await self._determine_shape_for_grab()
         self.uca.start_recording()
 
     @_translate_gerror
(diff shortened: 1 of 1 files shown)


MarcusZuber:
Together with https://github.com/ufo-kit/uca-net/commit/fe803befe5019971b1d7348443337eeb3ec68b0b it should work

MarcusZuber:
@tfarago On my pc at home it works. We should test this soon with real cameras. 
This will save me so many phone calls...

tfarago:
Why is the interface different in the first place? If that cannot be changed then we can proceed with this...

MarcusZuber:
The different interface has something to do where the command is executed (in the local net-plugin or in the remote plugin). The state is handled in the parent class of the net-plugin, which does not synchronize with the parent class of the server-side-plugin. This `do_` function explicitly call the function of the server-side and uses the state on the server.
I'm currently not deep enough in the uca and glib stuff to make this more elegant in uca-net.
If you don't want to change the uca-net code further I would proceed with this implemention.

tfarago:
alright :+1:

tfarago:
@MarcusZuber does this still work? Can I merge?

MarcusZuber:
I would like to do further tests with different hardware.
I would propose to postpone this after the python3 conversion.

tfarago:
Alright

tfarago:
This needs [libuca#44](https://github.com/ufo-kit/libuca/pull/44) because right now even `uca.is_recording` is not synced with the real remote camera.

MarcusZuber:
This looks so clean. Are you sure it works? :wink:

tfarago:
Now it should work.

tfarago:
And it also needs [this](https://github.com/ufo-kit/uca-net/pull/14/commits/7f9609633ecb6772ce24e2585142a1d633a6278a)

PR #410, opened 2018-05-29
Faster dummy camera

tfarago:
Currently it takes 800 ms to simulate 2560 x 2160 images which is unacceptable.

### DIFF
--- a/concert/tests/unit/devices/test_camera.py
+++ b/concert/tests/unit/devices/test_camera.py
@@ -11,7 +11,8 @@ class TestDummyCamera(TestCase):
 
     def setUp(self):
         super(TestDummyCamera, self).setUp()
-        self.camera = Camera()
+        self.background = np.ones((256, 256), dtype=np.uint16)
+        self.camera = Camera(background=self.background)
 
     def test_grab(self):
         frame = self.camera.grab()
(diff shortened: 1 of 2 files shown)


concert/tests/unit/devices/test_camera.py
```diff
@@ -72,6 +72,11 @@ def grab():
         image = self.camera.grab()
         np.testing.assert_equal(image, grab()[:, ::-1])
 
+    def test_simulate(self):
+        self.assertTrue(np.any(self.camera._background - self.camera.grab()))
```
tfarago: Accessing the underscored properties is not very clean, but is it acceptable in tests?

matze: I'd say no. In my opinion, a unit test should test the same interface the user will use at the end. In this case, I would generate a known background in the `setUp` step, pass it to the camera and save it in the test case object to assert it later in your `test_simulate` step.

tfarago: Maan, you like to make me work, don't you?! Naah you are right, now it should be good.

matze: It's just one more line ;-)

tfarago:
I squashed it already so it's good to go.

PR #411, opened 2018-06-01
FileCamera: Enable multi-page images

tfarago:
Not the most memory efficient solution, but no file-type-specific handling necessary.


matze:
LGTM :+1:

PR #416, opened 2018-10-02
Enable ctrl c

tfarago:
This stops all devices and experiments.

PR #417, opened 2018-10-02
Add progressbar

tfarago:
This is just fun to watch!

### DIFF
--- a/bin/concert
+++ b/bin/concert
@@ -7,6 +7,7 @@ import os
 import re
 import sys
 import shutil
+import signal
 import subprocess
 import tempfile
 import traceback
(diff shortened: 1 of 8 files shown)


tfarago:
Based on #416

bin/concert
```diff
@@ -395,6 +398,15 @@ class StartCommand(Command):
         # Add session path, so that sessions can import other sessions
         sys.path.append(cs.path())
 
+        # ctrl-c handling
+        orig_sigint_handler = signal.getsignal(signal.SIGINT)
```
tfarago: @matze the time I spent googling this is not so long, so do you perhaps know a cleaner way of doing this?

matze: I think this is alright. Signals are weird anyway, so …

PR #419, opened 2018-10-23
add fraction and percent to quantities

MarcusZuber:
This adds percent as a unit to the quantities. We need this for a device where all parameters are set in percent.

@tfarago: From my side this should be fine and can be merged if no one disagrees.


tfarago:
Can you give an example how you want to use this?

PR #431, opened 2019-12-17
Added ExternalLimitQuantity and ExternalLimitQuantityValue.

MarcusZuber:
This allows, that a Quantity can read additional limit values from a function. This can be used if a (physical) devices knows its limits and updates them internally.

### DIFF
--- a/concert/base.py
+++ b/concert/base.py
@@ -499,6 +499,29 @@ def quantity(unit=None, lower=None, upper=None, data=None, check=None, help=None
     return wrapper
 
 
+class ExternalLimitQuantity(Quantity):
+    """
+    A :class:`.Quantity` that reads external limits (e.g. given from a physical device, that knows  its current limits).
+    """
+    def __init__(self, unit, fget=None, fset=None, lower=None, upper=None, data=None, check=None, help=None,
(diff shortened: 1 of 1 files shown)


MarcusZuber:
@tfarago What do you think? After I finished, I readlized that the `ExternalLimitQuantity` shows the same behaviour as the `Quantity` (if the `limit_min_get` and` limit_max_get` are not set). If you like I could also put everithing directly in the `Quantity`.

tfarago:
I would add that to `Quantity` as you said. It's also how `State` works, it can use a function or a software value `_state_value`.

MarcusZuber:
I thought a little about this and decided to keep it separated, allowing the user to check the source of the current limit (see 1a617f1)

MarcusZuber:
Ok, to make everything kind of organized, I would propose, that an unset limit is `None` and no longer `(-)inf * unit`. Setting it to `inf` will still work.
However, I found here https://github.com/ufo-kit/concert/blob/master/concert/optimization.py#L67 one function in concert, where the limits are accessed directly (and don't handle a None).

Should I also change the `optimize_parameter` or should I try to find an other solution (I actual like the None for the unset values...)?

PR #434, opened 2019-12-17
Added State to Experiment

MarcusZuber:
Adds state to Experiment. Now running an already running experiemnt throws an exception.

### DIFF
--- a/concert/tests/integration/test_experiments.py
+++ b/concert/tests/integration/test_experiments.py
@@ -4,6 +4,9 @@ files creation.
 """
 import numpy as np
 import os.path as op
+import tempfile
+import shutil
+from time import sleep
 from concert.quantities import q
 from concert.coroutines.base import coroutine, inject
 from concert.coroutines.sinks import Accumulate
@@ -13,7 +16,7 @@ from concert.experiments.imaging import (tomo_angular_step, tomo_max_speed,
 from concert.experiments.addons import Addon, Consumer, ImageWriter, Accumulator
 from concert.devices.cameras.dummy import Camera
 from concert.tests import TestCase, suppressed_logging, assert_almost_equal, VisitChecker
-from concert.storage import DummyWalker
+from concert.storage import DummyWalker, DirectoryWalker
 
 
 class DummyAddon(Addon):
@@ -27,6 +30,25 @@ class DummyAddon(Addon):
     def _detach(self):
         self.attached_num_times -= 1
 
+
+class ExperimentSimple(Experiment):
+    def __init__(self, walker):
+        acq = Acquisition("test", self._run_test_acq)
+        super(ExperimentSimple, self).__init__([acq], walker)
+
+    def _run_test_acq(self):
+        for i in range(10):
+            sleep(0.1)
+            yield np.random.random((100, 100))
+
+
+class ExperimentException(Experiment):
+    def __init__(self, walker):
+        acq = Acquisition("test", self._run_test_acq)
+        super(ExperimentException, self).__init__([acq], walker)
+
+    def _run_test_acq(self):
+        raise Exception("Run test acq")
 
 @suppressed_logging
 def test_tomo_angular_step():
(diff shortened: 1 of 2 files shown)


tfarago:
Also, please test this.

MarcusZuber:
When using the `_state_value` it is quite ugly to check if an `abort()` finished (since the abort can not kill the future, but needs to be handled in the implementation of the frame gernarots). 
Therefore, I would prefer to check the state from the future as I did it in the first place to make sure, that really nothing is running in the background.

concert/tests/integration/test_experiments.py
```diff
@@ -242,3 +264,35 @@ def test_attach_num_times(self):
...
+        self.assertEqual(exp.state, "running")
+        exp.abort().join()
+        self.assertEqual(exp.state, "aborted")
+        exp_handle.join()
+
+    def test_experiment_exception(self):
+        exp = ExperimentException(self.walker)
+        self.assertEqual(exp.state, "standby")
+        exp_handle = exp.run()
+        sleep(0.1)
```
tfarago, line 297: Why do you need to sleep and not simply wait for `exp_handle.join()`?

tfarago:
> When using the `_state_value` it is quite ugly to check if an `abort()` finished (since the abort can not kill the future, but needs to be handled in the implementation of the frame gernarots).
> Therefore, I would prefer to check the state from the future as I did it in the first place to make sure, that really nothing is running in the background.

Why? in the end of `abort` you can set the state to abort. Until you reach that point in the code the state is anyway undefined because half the acquisitions might be getting aborted and the other half not yet. It's actually exactly what you are doing in the commit. Also, why do you sleep there? You like those sleeps don't you? :smile:

concert/tests/integration/test_experiments.py
MarcusZuber: join will block until the state is back to standby and the `_state_value` is not set fast enough. Without it you have a race condition if the state was set fast enough (that also why I don't like the `_state_value`)

MarcusZuber:
I hate those sleeps...
The commit with the abort works, if the user sets the state back from running. If this is not done correctly in the Acquisition the abort will block forever. And if I don't wait for the state change, it could be, that a 2nd run is started while the 1st one is still not completely aborted.
I still prefer the version with the checking of the future.

tfarago:
Let me play with this a little on Monday.

MarcusZuber:
What I could do, would be that I store the future in the class and check the state of the furture in the abort function and keep the state in "aborting" or something similar until it joins. But then, the difference of directly checking the future  state is quite little.

tfarago:
FYI:I had to force-push because master was ahead.

tfarago:
How about 60f479d5bd079e2103f3f78d150a633b2cd4a5fe? There is no way you can start before aborting is finished and I don't see how abort can block forever.

MarcusZuber:
`Acquisition.abort().join()` will stop the generation of new frames, but what happens with the `Acquisition.producer`? I guess, it will still run and needs to be somehow notified about the abort. 
Also, in https://github.com/ufo-kit/concert/commit/60f479d5bd079e2103f3f78d150a633b2cd4a5fe the state will not change from aborting to aborted (which maybe can be done after the `acq.abort()`)

tfarago:
> `Acquisition.abort().join()` will stop the generation of new frames, but what happens with the `Acquisition.producer`?

This is actually unrelated to this PR.

> the state will not change from aborting to aborted (which maybe can be done after the acq.abort())

Why should it? It's changed to `standby` automatically after `run` finishes. Do you really need to know about aborted state? Should the devices have `aborted` state? I would like to keep things consistent.

tfarago:
> `Acquisition.abort().join()` will stop the generation of new frames, but what happens with the `Acquisition.producer`?

AFAIK the producer coroutine finishes as well. If you put it's code in try-finally the finally clause gets executed.

MarcusZuber:
> Why should it? It's changed to `standby` automatically after `run` finishes. Do you really need to know about aborted state? Should the devices have `aborted` state? I would like to keep things consistent.

That's right.



> > `Acquisition.abort().join()` will stop the generation of new frames, but what happens with the `Acquisition.producer`?
> 
> This is actually unrelated to this PR.

Yes for this PR it is not so important. However, I think the abort should stop everything inside of the experiment and should somehow allow to check if it really stopped and not only stopping the generation of frames.



> > `Acquisition.abort().join()` will stop the generation of new frames, but what happens with the `Acquisition.producer`?
> 
> AFAIK the producer coroutine finishes as well. If you put it's code in try-finally the finally clause gets executed.

What do you think of splitting this in two functions? We could put this try-finally block in the Acquisition and the "user" provides a producerfunction (which will be in the try-part and can be the producer as it is used so far) and can provide an additional "abort-handler"-function which will be called in the finally-block?

tfarago:
> Yes for this PR it is not so important. However, I think the abort should stop everything inside of the experiment and should somehow allow to check if it really stopped and not only stopping the generation of frames.

Then let's finish what was the purpose of this PR here and start a clean new issue/PR elsewhere.

> What do you think of splitting this in two functions? We could put this try-finally block in the Acquisition and the "user" provides a producerfunction (which will be in the try-part and can be the producer as it is used so far) and can provide an additional "abort-handler"-function which will be called in the finally-block?

Isn't this what `finish` is for?

tfarago:
But yeah, that doesn't help on the finer acquisition level...

MarcusZuber:
Ok. Can you merge it with your last change? I will create a new PR with this abort-handler after you merged. The we can discuss this there.

Issue #363, opened 2015-02-13
Quantity that features measured value and target value


MarcusZuber:
I have a few cases in which I have a Quantity where I can set a target value and I can read a measured value, like in a Motor where I set the target position (then the controller tries to achieve the position) and I can read the real encoder position. 
In my most implementations the setter sets the target value and the getter reads the measured value. But this creates an (in my opinion) unwanted behavior, when using stash and restore.

My idea was to create a `ControlledQuantity` that features a target value (readable and writable) and a readable measured value. The stash/restore could then only restore the target values independently on any 'noisy' measured value.

Is there any elegant solution?

tfarago:
can you provide a use case for this? It will be easier to discuss on that.

MarcusZuber:
Ok. For example the x-ray tube.
`tube.set_voltage(120*q.kV)`
To reach the voltage setting it can take up to 1-2 minitues.
If I read `tube.voltage` in this time I will get someting inbetween the old voltage and the `120*q.kV` and if I use stash the unsetteled value will be stored. Also sometimes the controller is not able to set the selected value (like I'm always 1kV lower with the measured value). With using stash/resore multiple times I will deviate each time more from the target value.
Do you get my point?

tfarago:
Yes. Sounds reasonable and not too complicated to implement. I will come up with a PR at some point.

PR #435, opened 2020-01-11
Add external softlimits and taget values to Quantities

MarcusZuber:
Github became confused by my rebasing. This is a copy of https://github.com/ufo-kit/concert/pull/431

abb722d implements https://github.com/ufo-kit/concert/issues/363

### DIFF
--- a/concert/base.py
+++ b/concert/base.py
@@ -570,7 +570,8 @@ class ParameterValue(object):
     @property
     def writable(self):
         """Return True if the parameter is writable."""
-        return getattr(self._instance, '_set_' + self.name) is not _setter_not_implemented
+        return (getattr(self._instance, '_set_' + self.name) is not _setter_not_implemented or
+                self._parameter.fset is not None)
 
     @property
     def target_readable(self):
--- a/concert/base.py
+++ b/concert/base.py
@@ -425,7 +425,8 @@ class Quantity(Parameter):
     """A :class:`.Parameter` associated with a unit."""
 
     def __init__(self, unit, fget=None, fset=None, fget_target=None, lower=None, upper=None,
-                 data=None, check=None, help=None):
+                 data=None, check=None, external_lower_getter=None, external_upper_getter=None,
+                 help=None):
         """
         *fget*, *fset*, *data*, *check* and *help* are identical to the
         :class:`.Parameter` constructor arguments.
@@ -439,6 +440,8 @@ class Quantity(Parameter):
 
         self.upper = upper
         self.lower = lower
+        self.external_lower_getter = external_lower_getter
+        self.external_upper_getter = external_upper_getter
 
     def convert(self, value):
         if self.unit == "delta_degC":
@@ -690,6 +693,8 @@ class QuantityValue(ParameterValue):
         super(QuantityValue, self).__init__(instance, quantity)
         self._lower = quantity.lower
         self._upper = quantity.upper
+        self._external_upper_getter = quantity.external_upper_getter
+        self._external_lower_getter = quantity.external_lower_getter
         self._limits_locked = False
 
     def lock_limits(self, permanent=False):
@@ -707,7 +712,16 @@ class QuantityValue(ParameterValue):
 
     @property
     def lower(self):
-        return self._lower
+        if self.lower_user is None and self.lower_external is None:
+            return None
+        elif self.lower_user is None and self.lower_external is not None:
+            return self.lower_external
+        elif self.lower_user is not None and self.lower_external is None:
+            return self.lower_user
+        else:
+            return np.max((self.lower_user.to(self._parameter.unit).magnitude,
+                           self.lower_external.to(self._parameter.unit).magnitude),
+                          axis=0) * self._parameter.unit
 
     @lower.setter
     def lower(self, value):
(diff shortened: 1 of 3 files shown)


MarcusZuber:
abb722d implements #363: This means a Quantity can now have a `_get_targe_<name>` getter, that will be used by stash. The normal getter returns the measured value. This can be used if the measured value is not accurate enough and if a Device allows to read its target value. Then the restore resets the target value and not the noise measured value.

concert/base.py
```diff
@@ -662,9 +667,13 @@ def lower(self):
...
     @lower.setter
     def lower(self, value):
+        if value is None:
+            self._lower = None
+            return
         self._check_limit(value)
-        if value >= self._upper:
-            raise ValueError('Lower limit must be lower than upper')
+        if self._upper is not None:
+            if value >= self._upper:
```
tfarago: You could simply write ` if self._upper is not None and value >= self._upper:`

concert/base.py
```diff
@@ -673,9 +682,13 @@ def upper(self):
...
     @upper.setter
     def upper(self, value):
+        if value is None:
+            self._upper = None
+            return
         self._check_limit(value)
-        if value <= self._lower:
-            raise ValueError('Upper limit must be greater than lower')
+        if self._lower is not None:
+            if value <= self._lower:
```
tfarago: Same here

concert/optimization.py
```diff
@@ -64,10 +64,12 @@ def function(x_val):
         Create a function from setting a parameter and getting a feedback.
         """
         # Do not go out of the limits
-        if x_val < parameter.lower:
-            x_val = parameter.lower
-        if x_val > parameter.upper:
-            x_val = parameter.upper
+        if parameter.lower is not None:
+            if x_val < parameter.lower:
```
tfarago: same here

concert/optimization.py
```diff
@@ -64,10 +64,12 @@ def function(x_val):
...
+                x_val = parameter.lower
+        if parameter.upper is not None:
+            if x_val > parameter.upper:
```
tfarago: and here

concert/base.py
```diff
@@ -552,6 +567,10 @@ def writable(self):
         """Return True if the parameter is writable."""
         return getattr(self._instance, '_set_' + self.name) is not _setter_not_implemented
 
+    @property
+    def target_readable(self):
+        return getattr(self._instance, '_get_target_' + self.name) is not _getter_target_not_implemented
```
tfarago: Wouldn't this fail if I specified in the constructor e.g. `fget_target = 'foobarbaz'?

concert/base.py
```diff
@@ -597,7 +628,10 @@ def stash(self):
         if not self.writable:
             raise ParameterError("Parameter `{}' is not writable".format(self.name))
 
-        self._saved.append(self.get().result())
+        if self.target_readable:
+            self._saved.append(self.get_target().result())
+        else:
+            self._saved.append(self.get().result())
```
tfarago: Doesn't this discard what user sets? What if target sais -5 mm for the lower limit and I set it to -3 mm, then stash will not save what I specified, right?

concert/tests/unit/test_parameter.py
```diff
@@ -36,6 +36,24 @@ def _set_foo(self, value):
...
 
+class FooDeviceTagetValue(BaseDevice):
+    foo = Quantity(q.mm)
+
+    def __init__(self, value):
+        super(BaseDevice, self).__init__()
+        self._value = value
+
+    def _get_foo(self):
+        delta = np.random.randint(low=1, high=100) * q.mm
```
tfarago: Why do you need randomness?

concert/tests/unit/test_parameter.py
```diff
@@ -194,6 +212,17 @@ def test_saving(self):
...
             device['no_write'].restore().join()
 
+    def test_saving_with_target_value(self):
+        device = FooDeviceTagetValue(0 * q.mm)
+        device['foo'].upper = None
+        device['foo'].lower = None
+        device['foo'].stash().join()
+        self.assertEqual(device['foo'].target_readable, True)
+        device.foo = 1 * q.mm
+        self.assertEqual(device._value, 1 * q.mm)
```
tfarago: Please always test the user attributes, not the hidden ones. This will immediately force you to get rid of the random number above.
Also, please test the combination of user + external limits and the case that it's not implemented (the proper exception should be raised).

MarcusZuber:
> What does this test exactly? If I remove all the added code the tests pass as well. I meant in commit [6cc6468](https://github.com/ufo-kit/concert/commit/6cc6468b20fa93bb7ce27449524959d2564e7fae)

That the info_table can be created in each case. For some implementation of devices at the moment we have the case that exactly this does not work.

concert/base.py
MarcusZuber: Yes, you are right. At one point I had a problem where this did not work, but now it works and I will change it.

concert/base.py
MarcusZuber: I did not think about this, I just implemented it in the same ways as for the setter, which also does not work... (I will also include tests for the old stuff)

concert/tests/unit/test_parameter.py
MarcusZuber: I had the impression, that it is more realistic in this way, but a fixed value should also work.

concert/base.py
MarcusZuber: Not really. The get_target() should give the last value that was set. In the implementation this can stored in the Quantity or Device, or it can be read from the physical device (e.g. tango you can ask what the last set value for a parameter was).

tfarago:
Can you squash to essential commits and rebase? It's hard to review like this.

MarcusZuber:
Yes. This was my plan for the afternoon.

MarcusZuber:
@tfarago done

tfarago:
If you fix flake8 issues you can merge.

MarcusZuber:
Done

PR #436, opened 2020-03-24
Refactor align_rotation_axis

tfarago:


MarcusZuber:
Does this also work with a sphere instead of a needle tip (typically used for laminography)?

Could you also return the measured angles, that are not optimized?
E.g. in the laminography case I only want to optimize the roll, but it would be nice to get the measured pitch angle for the reconstruction.

tfarago:
> Does this also work with a sphere instead of a needle tip (typically used for laminography)?

Do you have data I could test it on?


> 
> Could you also return the measured angles, that are not optimized?
> E.g. in the laminography case I only want to optimize the roll, but it would be nice to get the measured pitch angle for the reconstruction.

Sure, will do.

MarcusZuber:
> > Does this also work with a sphere instead of a needle tip (typically used for laminography)?
> 
> Do you have data I could test it on?

Mathias will send you some data.

tfarago:
Finished. Now you can check the angles by the future result or like this:
```python
images = acquire_frames_360(camera, rot_motor, 10).result()
tips = find_needle_tips(images)
# ot
tips = find_sphere_centers(images)
rotation_axis(tips)
```

tfarago:
For future me: tested with the needle, the new very sharp needle and Ni spheres from @MathiasHurst

tfarago:
@MarcusZuber you used this at the beam line, right? Can we merge?

MarcusZuber:
We used it with the needles, but I think you wanted to make it also to work with the spheres.

PR #440, opened 2020-04-06
Split producer in frame_producer and abort_handler

MarcusZuber:
I didn't test it so far, just to show you the idea of it.

### DIFF
--- a/concert/experiments/base.py
+++ b/concert/experiments/base.py
@@ -33,9 +33,10 @@ class Acquisition(object):
 
     """
 
-    def __init__(self, name, producer, consumers=None, acquire=None):
+    def __init__(self, name, producer, abort_handler=None, consumers=None, acquire=None):
         self.name = name
-        self.producer = producer
+        self._frame_producer = producer
+        self._abort_handler = abort_handler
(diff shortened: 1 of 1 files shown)


MarcusZuber:
In the abort_handler the user can disable everything, that maybe is still ruining from the producer.

tfarago:
This change seems weird to me, I as a generator produce something and if something goes wrong I tell the acquisition to come back to me for clean up. Why would I go through acquisition?

I personally would write my producer in such a way that it has no side effects, i.e. I would use try-finally inside it. (This has to be properly documented so that everyone who works with this knows they have to do it like this, which is currently not the case unfortunately...) I think that's your motivation behind it, right? But let's just not forget that there _must_ be some responsibility also left on the programmers who write the executive code.

tfarago:
for example:

```python
def produce():
    try:
        for i in range(10):
            yield camera.grab()
    finally:
        self.stop_motors().join()

acq = Acquisition('radios', produce)
```

tfarago:
Pin this, possibly re-visit after python 3 transition and native asyncio support.

PR #442, opened 2020-04-09
Execute finish when an exception was raised in an acquisition

MarcusZuber:
@farago After our discussion about the state and some discussions what happened at the beamline, I realized that finish will not be executed if an exception is raised in an Acquisition. This commit will always execute the finish.

### DIFF
--- a/concert/experiments/base.py
+++ b/concert/experiments/base.py
@@ -233,17 +233,23 @@ class Experiment(Parameterizable):
         try:
             self.prepare()
             self.acquire()
-            self.finish()
         except:
             self._state_value = 'error'
             LOG.exception('Error while running experiment')
             raise
         finally:
-            if self.separate_scans and self.walker:
-                self.walker.ascend()
-            LOG.debug('Experiment iteration %d duration: %.2f s',
-                      self.iteration, time.time() - start_time)
-            self.iteration += 1
+            try:
+                self.finish()
+            except:
+                self._state_value = 'error'
+                LOG.exception('Error while running experiment')
+                raise
+            finally:
+                if self.separate_scans and self.walker:
+                    self.walker.ascend()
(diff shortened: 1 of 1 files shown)


MarcusZuber:
I just realized that an exception in the finish  can now screw up the walker. I will make this more stable.

MarcusZuber:
This should do the trick.

tfarago:
Why did you take the rest of the finally block out?

MarcusZuber:
Then an exception in the `finish()` does not stop the walker-actions.

tfarago:
But if there is an exception in the `run` method then the walker never ascends, I think the highest priority is the `finish`, so I would put that first in the `finally` block and then the walker stuff behind, but still in the `finally` block.

MarcusZuber:
walker.descend ist the only line, that could cause an exception, that is not handled, and it this fails we do not need the walker ascend afterwards.

MarcusZuber:
Oh, sorry, I missed the raise in the except block. You are right. I will put it back in the finally block.

MarcusZuber:
@tfarago what do you think now?

tfarago:
Did you test the nested try-finally? If it works you can merge, I see why you'd do it like this...

MarcusZuber:
I tested it. 

The only thing, that is not 200% perfect is the case, when you have an exception in the acquisitions and in the `finish()`. Then the future from the `run()` gives you only the exception in the finish. However, in the log all exceptions are written.
I could live with that (that all exceptions can not be handled automatically from something that runs the experiments). Also, in such a case something is seriously broken and maybe should be checked manually.

If you agree, we can merge it.

tfarago:
Let's just test it in practice and we can still come back to this later if it's too confusing with the exceptions.

PR #445, opened 2020-04-30
experiments: add per-run log handler

### LINKED ISSUE #438: Seperate log files for single experiment executions
I often put something like this in the `Experiment.prepare` to get a logger in the Experiment class, that creates one (additional) log-file for each experiment run in the folder of the current walker position, that has only the experiment related infos.

```
def prepare(self):
	self.logger = logging.getLogger("CTScan")
	handler = logging.FileHandler(self.walker.current + "/CTScan.log")
	self.logger.handlers = []

	formatter = logging.Formatter('%(asctime)s - %(name)s - %(levelname)s - %(message)s')
	handler.setFormatter(formatter)
	handler.setLevel(logging.INFO)
	self.logger.addHandler(handler)

	self.logger.info("Exposure time %s" % self._camera.exposure_time)
```

I could add this directly in the base Experiment class, if nobody sees a drawback by doing this.

tfarago:
fixes #438

### DIFF
--- a/concert/experiments/base.py
+++ b/concert/experiments/base.py
@@ -4,10 +4,11 @@ care of proper logging structure.
 """
 
 import logging
+import os
 import time
 from concert.async import async
 from concert.progressbar import wrap_iterable
-from concert.base import Parameterizable, Parameter, State, check
+from concert.base import Parameterizable, Parameter, Selection, State, check
 
 
 LOG = logging.getLogger(__name__)
@@ -104,6 +105,7 @@ class Experiment(Parameterizable):
     separate_scans = Parameter()
     name_fmt = Parameter()
     state = State(default='standby')
+    log_level = Selection(['critical', 'error', 'warning', 'info', 'debug'])
 
     def __init__(self, acquisitions, walker=None, separate_scans=True, name_fmt='scan_{:>04}'):
         self._acquisitions = []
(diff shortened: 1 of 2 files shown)


MarcusZuber:
Ha, almost the same as I just did ;-)

MarcusZuber:
The only difference it the llog-level. What do you think? Do we want to have the global log level, or an experiment specific?
I would prefer a experiment specific one, so I don't get all the spam of the motor setter etc. when I test a new Experiment.

MarcusZuber:
You should also chane the LOG in the experiment to self.log

tfarago:
> Ha, almost the same as I just did ;-)

Did we really push this at the same time after those weeks of nothingness??

MarcusZuber:
> > Ha, almost the same as I just did ;-)
> 
> Did we really push this at the same time after those weeks of nothingness??

I waited until the other experiment stuff was finished.

tfarago:
> The only difference it the log-level. What do you think? Do we want to have the global log level, or an experiment specific?

We can add that.

> I would prefer a experiment specific one, so I don't get all the spam of the motor setter etc. when I test a new Experiment.

There is no clutter thanks to the logging hierarchy.

MarcusZuber:
Ah, one more thing. We could automatically print all Parameters in the log file in the `run()`. To make it a little bit more "idiot proof".

MarcusZuber:
Just with 
`self.log.info(self)`

tfarago:
Done and done, please check.

tfarago:
I like LOG because it's consistent. what users do in their own code is their business.

PR #447, opened 2020-04-30
Add device names

tfarago:
which is convenient for logging.

### DIFF
--- a/concert/base.py
+++ b/concert/base.py
@@ -50,6 +50,39 @@ def _execute_func(func, instance, *args, **kwargs):
     return result
 
 
+def _find_object_by_name(instance):
+    """Find variable name by *instance*. This is supposed to be used only in the
+    :meth:`.Parameter.__set__`.
+    """
+    def _find_in_dict(dictionary):
+        for (obj_name, obj) in dictionary.items():
+            if obj == instance:
+                return obj_name
+
+    instance_name = None
+
+    frames = inspect.stack()
+    try:
+        # Skip us and Parameter.__set__
+        for i in range(2, len(frames)):
+            # First look in the globals
+            instance_name = _find_in_dict(frames[i][0].f_globals)
+            if not instance_name:
+                # If not found, look in the locals (e.g. devices instantiated in funcions)
+                instance_name = _find_in_dict(frames[i][0].f_locals)
+            # We can't know at which index of the stack is the correct name, e.g. if a device sets a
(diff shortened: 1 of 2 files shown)


tfarago:
Using Tango names and so on is trivial.

MarcusZuber:
Is it possible to get the name of the instance and put it as the name in the class? `ddoc()` also somehow figures out the names in the global scope.

tfarago:
> What do you think of moving this to `Parameterizable`? Then also experiments would have names.

That's a very abstract thing, I wouldn't put it there

tfarago:
> Is it possible to get the name of the instance and put it as the name in the class? `ddoc()` also somehow figures out the names in the global scope.

Unfortunately it doesn't work for devices instantiated in the session. Apart from that, it's super-magicky and very slow. Just asking for the stack takes 3.5 ms on my machine.

I will play with it anyway and we can decide later.

tfarago:
19ff89c966401d23ddccbcdd9a7e3c1c96b35a42 adds what you asked for.

tfarago:
@MarcusZuber 13ae24a  implements the magicky name resolution in a more efficient way, it looks up the name only once and then caches thanks to `memoize`. I can live with this for now, at least we could test it. IMO convenience wins over magic.

tfarago:
So I'd remove all the commits before.

MarcusZuber:
When the device is declared in the session it works:
`[2020-05-07 11:59:07,577] INFO: concert.base: test: set Camera::dimax.frame_rate='100.0 / second'
`
When created within the running session this is the result:

`[2020-05-07 12:00:12,720] INFO: concert.base: test: set Camera::self.frame_rate='5.0 / second'
`

I think this is fine so far.

tfarago:
Alright, now it tries to be even more clever. Please check again and if there are no complaints, I will update the documentation and merge.

MarcusZuber:
I would like to try it under real conditions on monday again. I saw a strange interference with our local sessions, that I don't understand so far.

tfarago:
Alright, we wait. Can you describe the interference in the meantime? Maybe I can figure something out even before Monday.

MarcusZuber:
Now it works...
Maybe I was just too tired yesterday.

The only thing I just found out. In the log the name is written, however the name-property of the device is not automatically set. What is the difference between `Device.name` and `Property.name_for_log` ?

tfarago:
There is no name property, I will discard the old commits.

tfarago:
Alright, let's give this a try.

PR #455, opened 2021-04-13
Python 3 conversion

tfarago:
### Coroutines and Asynchronous Execution

- Similar to how `gevent` works, we will use native coroutines to execute things _concurrently_ and not stuff everything into the `ThreadPoolExecutor`, since asyncio does basically the same as we want, `gevent` will be removed. 
- `@coroutine` will be replaced by native `async def` coroutines.
- Concert core will by fully based on native coroutines ~and blocking functions for user convenience will be created outside of the core, probably in `concert.commands` module which will implement functionality for session convenience functions via `@command` decorator and `create_command` function for dynamic shortcuts creation.~ (nope, this was weird)
- In general, based on whether code starts right away and whether it blocks the command line, there will be ~three~ two cases:
  - *non-blocking* **background tasks**: `task = start(corofunc())`, most of the `async def` functions close to the user interface are decorated with `@background`, which does exactly the same as `start(corofunc())`
  - *blocking* **foreground tasks**: `await corofunc()`
  - ~*started*, *blocking* **commands**: `cmd()`~ (removed because of the confusing syntax)

- For device parameters, things will look like this:
  - `task = dev.get_parameter()`, _non-blocking_, this wraps the coroutine into a task and runs it in the background, use case: programmers or more complex session usage.
  - `result = dev.parameter` _blocking_, returns parameter value, for usage in concert session

#### Notes
According to [this](https://ipython.readthedocs.io/en/stable/interactive/autoawait.html) and [this](https://github.com/ipython/ipython/blob/master/IPython/terminal/pt_inputhooks/asyncio.py), one can use IPython's REPL in combination with `%gui asyncio` to let coroutine tasks run in the background, similarly to how the REPL in Python 3.8 `python -m asyncio` described [here](https://docs.python.org/3/whatsnew/3.8.html) works. In Python, one can let the task run in the background by invoking `asyncio.create_task(coro_func())`. In IPython one does that by `asyncio.ensure_future(coro_func())`

Cancellation behavior:

- `ctrl-c` in a blocking call (e.g. `time.sleep()`): running task not cancelled and we get a `KeyboardInterrupt` in a running coroutine 
- `ctrl-c` in an async call (e.g. `asyncio.sleep()`): running task cancelled and we get a `CancelledError` in a running coroutine (but only if we implement the cancellation via `SIGINT`, otherwise `%gui asyncio` doesn't cancel the task

Example of a device:
```python
from concert.coroutines.base import background, start

class Device:
    value = Parameter()

    async def _set_value(self, value):
        await network.send(value)

    @background
    async def set_value(self, value):
        await self._set_value(value)

    @background
    async def do_something(self):
        pass

d = Device()
await d.set_value(123) # block
d.value = 123 # block
f = d.set_value(123) # do not block (run in background), f is asyncio.Future and can be awaited
d.do_something() # run in background
await d.do_something() # block

d_2 = Device()
# set values concurrently and block until they finish
await asyncio.gather(d.set_value(123), d_2.set_value(321))
# set values concurrently and do not block (run in background)
f = start(asyncio.gather(d.set_value(123), d_2.set_value(321)))
```
Implement coroutine functions and asynchronous generators explicitly, so that it is easy to check with `inspect.iscoroutinefunction` and `inspect.isasyncgenfunction`, which may be useful tools for users who will dynamically want to check if a function is a coroutine or not. Example:
```python
async def asyncgen():
    for i in range(5):
        yield i

# Although a valid way to create the async generator, do not do it like this:
def middleman_impl():
    return asyncgen()

# Rather like this:
async def middleman_expl():
    async for item in asyncgen():
        yield item

# Because
print(inspect.isasyncgenfunction(middleman_impl))
False
print(inspect.isasyncgenfunction(middleman_expl))
True
```

### Viewers, Loops, Signals

Currently, `%gui asyncio` is not combinable with matplotlib. We could provide a custom input hook like this:
```python
def mpl_inputhook(context):
    import matplotlib.pyplot as plt

    while not context.input_is_ready():
            plt.pause(0.01)

IPython.terminal.pt_inputhooks.register('ih', mpl_inputhook)
shell.enable_gui(gui='ih')
```
but then there would be a [separate loop created for our shell](https://github.com/ipython/ipython/blob/master/IPython/terminal/interactiveshell.py#L651), which effectively means that user coroutines couldn't run in the background (unless we forced the loop in `concert.coroutines.base.start` to be the `pt_loop` of our shell, but that leads to all kinds of trouble related to loop switching). The lesser evil seems to be to have a non-cancellable non-keyboardinterruptable coroutine doing basically the same as the `mpl_inputhook` function above. Right now we need to focus on porting the rest of the code, this can be cleaned up later. However, the safest way is probably to just render everything in a **separate process** like before, so we will stick with that. On the top of that, instead of rendering being the "heavy-lifting" part, with multiprocessing it's the `queue.put`, which is the same for all backends.

**Never ever import matplotlib in the session if using multiprocessing-based drawing. Somehow the matplotlib gets screwed up and for example a second viewer will not open.**

https://matplotlib.org/stable/users/interactive_guide.html
https://matplotlib.org/stable/gallery/misc/multiprocess_sgskip.html
https://ipython.readthedocs.io/en/stable/config/eventloops.html
https://vorpus.org/blog/control-c-handling-in-python-and-trio/

### Inspiration:

https://github.com/timofurrer/awesome-asyncio
https://github.com/vxgmichel/aiostream
https://stackoverflow.com/questions/50039223/how-to-execute-two-aggregate-functions-like-sum-concurrently-feeding-them-f


bin/concert
```diff
@@ -65,8 +65,8 @@ def get_ipython_shell(config=None):
             allow_CTRL_C()
             inputhook_manager._current_gui = 'gevent'
         else:
-            print("IPython {} is too old for gevent integration. "
-                  "Please upgrade IPython or disable gevent.".format(version))
+            print(("IPython {} is too old for gevent integration. "
```
MarcusZuber, line 68: This 2to3 is so sophisticated...

concert/experiments/addons.py
```diff
@@ -137,14 +137,14 @@ class ImageWriter(Addon):
 
     A :class:`~concert.storage.Walker` instance
 
-    .. py:attribute:: async
+    .. py:attribute:: casync
```
MarcusZuber, line 140: It does not break it, but I think here happend a little bit to much find-and-replace.

tfarago: Yeah I will polish these things.

bin/concert
tfarago: Yep, I am gonna have some fun with `sed` in some cases.

tfarago:
Does anybody know how to do broadcasting with native Python 3 coroutines?? Perhaps @matze has some insights and may spare me the googling...

tfarago (commented):
I am not sure if it's a good idea to mix new- and old-style approach of chaining coroutines but we need to be able to broadcast. Perhaps there are some libraries out there which can do this (and more) elegantly, need to check that.

tfarago:
https://github.com/timofurrer/awesome-asyncio

tfarago (commented):
Don't forget to clean up the threadpoolexecutor. Also, what to do with dispatcher?

MarcusZuber:
`concert.commands.COMMANDS`, `concert.helpers.Command` and the command-decorator confused me a little bit. Maybe renaming the helpers.Command to something else would make this more cleaned-up.

tfarago:
> `concert.commands.COMMANDS`, `concert.helpers.Command` and the command-decorator confused me a little bit. Maybe renaming the helpers.Command to something else would make this more cleaned-up.

Yep, that's right. I cleaned it up in 5e43b8605c0f3ac55be81cef13fd8693ba9ac8bc. If there are no more remarks, I would slowly start squashing and merge. The change is so big that we will have to fine-tune it in real environments later anyway.

PR #457, opened 2021-12-17
New experiments

MarcusZuber:
Added typical experiments for x-ray tube based labs and synchrotron.

PR #458, opened 2021-12-18
New experiments

MarcusZuber:
Added typical experiments for x-ray tube based labs and synchrotron.

Comment about new PW:
I accidently deleted the branch here when rebasing and force pushing here https://github.com/ufo-kit/concert/pull/457.
Github refused to open the PR after re-uploading.

Todo:
- [x] write tests
- [x] speedup test
- [x] make flake8 happy
- [x] write docs

### DIFF
--- a/concert/experiments/imaging.py
+++ b/concert/experiments/imaging.py
@@ -1,6 +1,13 @@
-"""Imaging experiments usually conducted at synchrotrons."""
+"""Imaging experiments.
+
+For all classes the abstract functions *start_sample_exposure* and *stop_sample_exposure* need to be
+ implemented."""
+import asyncio
 import numpy as np
+import concert.storage
 from concert.quantities import q
+from concert.experiments.base import Experiment, Acquisition
+from concert.base import background, Parameter, Quantity, Parameterizable
 
 
 async def frames(num_frames, camera, callback=None):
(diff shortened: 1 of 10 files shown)


concert/experiments/imaging.py
```diff
@@ -60,3 +63,555 @@ def tomo_max_speed(frame_width, frame_rate):
...
+            await after_acquisition()
+
+    async def start_sample_exposure(self):
+        raise NotImplementedError()
+
+    async def stop_sample_exposure(self):
+        raise NotImplementedError()
+
+
+class SteppedTomography(Radiography):
```
tfarago, line 198: Shouldn't we call this just `Tomography`. `ContinuousTomography` is for me the opposite of stepped and here it inherits...

MarcusZuber: To make it really correct, from the inheritance, nameing etc. I could do
imaging

MarcusZuber: And then also add SteppedSpiralTomo, where I need to think a little bit to not add to much code duplication.

tfarago: Yes, but if `SteppedTomography` turns out to be just empty then you can just omit it.

MarcusZuber: Tomography will be almost emplty. Only the rot-motor and some parameters will be introduced. Then the thw tomo-implementations will implement the prepare_radios and take_radios.
I think this is the most clear and most educational solution.

concert/experiments/imaging.py
```diff
@@ -60,3 +63,555 @@ def tomo_max_speed(frame_width, frame_rate):
...
+    async def _set_num_projections(self, n):
+        self._num_projections = int(n)
+
+    async def _set_flat_position(self, position):
+        self._flat_position = position
+
+    async def _set_radio_position(self, position):
+        self._radio_position = position
+
+    async def _prepare_flats(self):
```
tfarago, line 127: Is there a reason why these should be private?

concert/experiments/imaging.py
```diff
@@ -327,45 +343,22 @@ async def stop_sample_exposure(self):
...
 
-class SpiralTomography(ContinuousTomography):
-    """
-    Spiral Tomography
-
-    This implements a helical acquisition scheme, where the sample is translated perpendicular to the beam while the
-    sample is rotated and the projections are recorded.
-    """
-    vertical_velocity = Quantity(q.mm / q.s)
+class Spiral(Parameterizable):
```
tfarago, line 346: Shouldn't this be called `SpiralMixin` actually?

concert/experiments/imaging.py
```diff
@@ -387,9 +380,102 @@ async def _set_sample_height(self, height):
...
+            self._vertical_motor.set_position(await self.get_start_position_vertical())
+        )
+
+    async def _finish_radios(self):
+        if self._finished:
+            return
+        await self._prepare_darks()
+        await self._tomography_motor.stop()
+        await self._vertical_motor.stop()
+        await self._vertical_motor['motion_velocity'].restore()
```
tfarago, line 411: Couldn't all these be executed concurrently and gathered like the line below?

concert/experiments/imaging.py
```diff
@@ -60,3 +63,555 @@ def tomo_max_speed(frame_width, frame_rate):
...
+            self._vertical_motor.set_position(await self.get_start_position_vertical())
+        )
+        self._finished = True
+
+    async def _take_radios(self):
+        try:
+            await self._prepare_radios()
+            await self._vertical_motor['motion_velocity'].stash()
+            await self._tomography_motor.set_velocity(await self.get_velocity())
+            await self._vertical_motor.set_velocity(await self.get_vertical_velocity())
```
tfarago: Also `gather`?

concert/experiments/imaging.py
MarcusZuber: The stops yes. Changing the Motion Velocity while the Motor is moving is forbidden or could result in strange behaviour.

concert/experiments/imaging.py
MarcusZuber: Yes

concert/experiments/imaging.py
MarcusZuber: The property setter is always private. 

I thought the prepare/finish- functions would be only called from within the Experiment class. 
If you they should be called by the User, I can make them public.

concert/experiments/imaging.py
MarcusZuber: Yes

tfarago:
All and all I like this a lot. Once flake8 is happy and we have tests this can be merged.

MarcusZuber:
Docs and tests are still in progress, the Experiments seam to be finished (lets wait for the tests...).

concert/experiments/imaging.py
```diff
@@ -60,3 +66,451 @@ def tomo_max_speed(frame_width, frame_rate):
...
+        await self._tomography_motor.stop()
+        await self._vertical_motor.stop()
+        await self._vertical_motor['motion_velocity'].restore()
+        await asyncio.gather(
+            self._tomography_motor.set_position(await self.get_start_angle()),
+            self._vertical_motor.set_position(await self.get_start_position_vertical())
+        )
+        self._finished = True
+
+    async def _take_radios(self):
```
tfarago, line 504: Apart from the vertical motion, this is the same as in the continuous tomography case, should be also refactored.

concert/tests/integration/test_imaging_experiments.py
```diff
@@ -324,34 +405,68 @@ def setUp(self):
                                                       num_darks=2,
                                                       num_projections=15)
         self.writer = ImageWriter(walker=self.walker, acquisitions=self.exp.acquisitions)
+        self.acc = Accumulator(self.exp.acquisitions)
+
+    async def test_number_of_frames(self):
+        if (await self.exp.get_iteration()) < 2:
```
tfarago, line 411: What is this for?

MarcusZuber: This is no longer there. Here I tried to convince my local test setup to run the experiment only once.

MarcusZuber:
Tests are done.

concert/experiments/imaging.py
tfarago: I think the comment is on `_prepare_flats`.

MarcusZuber: Do we want to have them public?

tfarago:
Wow, that's a lot of docs! I like it. However, there seems to be a conflict.

tfarago:
You merged, not rebased, which would have made a cleaner history. Please check my [rebased branch](https://github.com/ufo-kit/concert/tree/new-experiments2) and if all is fine I will force-push here and merge.

MarcusZuber:
> You merged, not rebased, which would have made a cleaner history. Please check my [rebased branch](https://github.com/ufo-kit/concert/tree/new-experiments2) and if all is fine I will force-push here and merge.

Just realized myself. I thought the "resolve" button in github would do a rebase not a merge...
I just rebased and force-pushed here. So if you are fine with it we can merge this.

Issue #454, opened 2021-03-18
Abstract class to run multiple (simmilar) experiments


MarcusZuber:
Having a base class for organizing loops over experiment executions (like scanning multiple samples or scanning one sample with different parameters) would be nice.
It would hold an experiment instance and can run it multiple times with the possibility to adjust parameters in between the executions.
@tfarago : In principle base class for the "director" we use in our sessions.

tfarago:
Yep, yep, yep, after python 3 switch.

MarcusZuber:
I put a simple implementation in https://github.com/ufo-kit/concert/pull/459.

My idea was, that the functions `_prepare_run(iteration: int)`,` _get_number_of_iterations()` and `_get_iteration_name(iteration: int)`
would be implemented in  an implementation of this base class.
Then the thing iterates up to the number_of_iterations and before the experiment is executed the` _prepare_run` does change settings or does something else to prepare the next execution of the experiment.

tfarago:
Then there should be also `_finish`, or? In the expriment we provide before and after flexibility, so it would make sense here too I guess.

MarcusZuber:
Implemented in https://github.com/ufo-kit/concert/pull/459 .

PR #459, opened 2021-12-20
General Experiment Director

MarcusZuber:
So far this is only intended as a basis for discussions how we would like to implement something, that runs experiments multiple times with different settings or specimens.
 
See https://github.com/ufo-kit/concert/issues/454


concert/processes/experiments.py
```diff
@@ -40,9 +40,9 @@ async def run(self):
 
             # prepare first iteration
             self._iteration = 0
-            await self._prepare_run(self._iteration)
 
             for iteration in range(await self.get_number_of_iterations()):
+                await self._prepare_run(self._iteration)
```
tfarago, line 45: I guess no point in having it twice.

tfarago: Oh wait, did you want `_prepare_next_run`? Anyway, this seems very coarse, I won't interfere until all your ideas are in properly.

concert/processes/experiments.py
```diff
@@ -86,9 +87,8 @@ async def run(self):
                                           await self._get_iteration_name(iteration))
 
                 if hasattr(self._experiment, "ready_to_prepare_next_sample"):
-                    while not self._experiment.ready_to_prepare_next_sample:
-                        await asyncio.sleep(0.1)
```
tfarago: yay!

tfarago:
Good idea.

MarcusZuber:
Here the example:
https://github.com/ufo-kit/concert-examples/blob/experiment-director/experiment-director.py

@tfarago I didn't test it (this and the example), so there can be some typos etc.. But you can already have a look to get the idea, since I will not have the time to proceed with this until friday.

concert/processes/experiments.py
```diff
@@ -49,12 +80,23 @@ async def run(self):
 
                 self._experiment.walker.descend(await self._get_iteration_name(iteration))
                 exp_run = self._experiment.run()
-                self._experiment.log.info("Sample name: %s" % await self._get_iteration_name(iteration))
+                self._experiment.log.info("Sample name: %s" %
+                                          await self._get_iteration_name(iteration))
+
+                if hasattr(self._experiment, "ready_to_prepare_next_sample"):
```
tfarago, line 89: Wouldn't it be worth making something like a signal in pyqt and registering for it here? I can imagine that by _every_ experiment one can define a state when no devices will be touched anymore and for experiments which do not know/care the default implementation can be conservative and set this event at the end of the run. The reason is that I want to move away from these ad-hoc "hasattr" checks, which stem from our "workarounds" back then,

tfarago: instead of a signal, asyncio.Event would work in this case.

MarcusZuber: Yes. I will add this to the base-Experiment and set/reset in the run(), so it will work if the subclass does not implement it.

MarcusZuber:
@tfarago : What do you think? If you agree about the general interface/behavior, I will clean add the tests and document it. I would like the finish this soon.

concert/processes/experiments.py
```diff
@@ -0,0 +1,126 @@
...
+
+        except asyncio.CancelledError:
+            # This is normal, no special state needed -> standby
+            LOG.warning('Experiment director cancelled')
+            self._state_value = 'standby'
+        except Exception as e:
+            # Something bad happened, and we can't know what, so set the state to error
+            LOG.warning(f"Error `{e}' while running experiment director")
+            raise StateError('error', msg=str(e))
+        except KeyboardInterrupt:
```
tfarago, line 103: This you don't need to handle, when it's async def it will be caught by session utils.

tfarago (changes requested):
Looks already quite usable.

concert/processes/experiments.py
MarcusZuber: Are you 100% sure? I have the impression, that this is what Daniel saw in his Experiment class.

tfarago: Right, rather keep it in, sorry.

concert/processes/experiments.py
MarcusZuber: I think it should be correct now.

concert/experiments/base.py
```diff
@@ -95,13 +95,20 @@ class Experiment(Parameterizable):
...
+        still running.
+
     """
 
     iteration = Parameter()
     separate_scans = Parameter()
     name_fmt = Parameter()
     state = State(default='standby')
     log_level = Selection(['critical', 'error', 'warning', 'info', 'debug'])
+    ready_to_prepare_next_sample = asyncio.Event()
```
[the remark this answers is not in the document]
MarcusZuber: Thx!

MarcusZuber:
Any idea what to test? I would add a test for checking the state after a broken and working experiment.

Should I leave in the xy-scan? I thought it's a good example and I also use it for laminography scans.

concert/helpers.py
```diff
@@ -232,3 +234,59 @@ def is_iterable(item):
         return True
     except Exception:
         return False
+
+
+def arange(start, stop, step):
```
tfarago, line 239: Maybe you can name it so that it's clear the stripping occurs...

tfarago:
> Should I leave in the xy-scan? I thought it's a good example and I also use it for laminography scans.

Sure

> Any idea what to test? 

Make sure the event works, callbacks, state transitions...

concert/helpers.py
MarcusZuber: But it strips the unit only internally. From the outside it looks like the numpy function but with proper unit handling. I think if one combination of numpy and pint this was possible a while ago.

MarcusZuber:
Tests are still ugly with a lot of code duplication, but it now tests the time of the prepare_next_sample (in the director) with and without the setting of the ready_to_prepare_next_sample. It also tests the state after a "good" and an experiment that raises an Exception.

MarcusZuber:
@tfarago Besides the uncleaned tests, what do you think about the general interface/functionality?

concert/directors/base.py
```diff
@@ -0,0 +1,151 @@
...
+
+    async def _get_current_iteration(self) -> int:
+        return self._iteration
+
+    @background
+    async def pause(self):
+        self._run_event.clear()
+
+    @background
+    async def unpause(self):
```
tfarago, line 144: ehm, you mean resume?

concert/directors/base.py
```diff
@@ -0,0 +1,151 @@
...
+                formatter = logging.Formatter('%(asctime)s - %(name)s - %(levelname)s '
+                                              '- %(message)s')
+                handler.setFormatter(formatter)
+                self.log.addHandler(handler)
+            self.log.info(await self.info_table)
+
+            await self.prepare()
+            # prepare first iteration
+            self._iteration = 0
+            await self._prepare_run(self._iteration)
```
tfarago, line 96: or `self._iteration = -1` and use the `_prepare_next_run` like for the rest of the iteration?

MarcusZuber:
I just realized, that the iteration property of the experiments starts at 1 and not at zero. Should we also make the Director iterations starting at 1? Or change the experiment iteration to start at zero (which I would prefer).

tfarago:
As you like.

concert/directors/base.py
MarcusZuber: Yes

concert/directors/base.py
MarcusZuber: I like it more this way. I thinks it is better to understand than calling `_prepare_run(0 - 1 + 1)`.

MarcusZuber:
Any additional suggestions regarding the tests?

tfarago:
Looks quite complete.

MarcusZuber:
Ok. I cleaned up the commits. Could you have a last look? It could be that there are still some pep8 things, this I didn't check, but I will before merge.

tfarago (approved):

tfarago:
But make sure the tests run before you hit the button!

MarcusZuber:
I found a really stupid error. I will fix it after async-all PR is done.

PR #461, opened 2022-01-17
Start making limits async

tfarago:
because some devices use async functions to get external limits.


concert/tests/integration/test_dummyfocusing.py
```diff
@@ -13,8 +14,8 @@ class TestDummyFocusingWithSoftLimits(TestCase):
     def setUp(self):
         super(TestDummyFocusingWithSoftLimits, self).setUp()
         self.motor = LinearMotor(position=50 * q.mm)
-        self.motor['position'].lower = 25 * q.mm
-        self.motor['position'].upper = 75 * q.mm
+        run_in_loop_thread_blocking(ensure_coroutine(self.motor['position'].set_lower, 25 * q.mm))
```
MarcusZuber, line 17: why do we need the thread loop here?

MarcusZuber (approved):

tfarago:
Did we seriously not have limits documented??? Shame on us... @MarcusZuber if what I wrote makes sense we can merge.

concert/tests/integration/test_dummyfocusing.py
tfarago: Of course we don't.

tfarago:
Now also the `run_in_loop_thread_blocking` was removed, should be done.

Issue #407, opened 2017-12-21
PyplotViewer coroutine not updating


MarcusZuber:
The PyplotViewer is not updating the plot when new data arrives through the coroutine.

```python
import numpy as np
from concert.helpers import Region
from concert.processes.common import scan
from concert.devices.motors.dummy import LinearMotor
from concert.async import resolve
from concert.coroutines.base import inject
from concert.ext.viewers import PyplotViewer

motor = LinearMotor()

def get_flow_rate():
    from time import sleep
    sleep(1)
    return 1.0

region = Region(motor['position'], np.linspace(5, 12, 10) * q.mm)
viewer = PyplotViewer()

inject(resolve(generator), viewer())
```

This updates the plot only sometimes. Also resizing the plot-window does not update the window.

Matplotlib: 1.5.1

MarcusZuber:
This is no longer present after the python3 conversion.

PR #464, opened 2022-01-28
Monochromator implementation with the possibility to scan

MarcusZuber:
Todo:
- [x] add tests
- [x] Test viewer (= check https://github.com/ufo-kit/concert/issues/407)

### DIFF
--- a/concert/devices/monochromators/doublemonochromator.py
+++ b/concert/devices/monochromators/doublemonochromator.py
@@ -1,0 +1,117 @@
+"""
+Module to implement a monochromator with a second crystal/multilayer that can be scanned.
+"""
+import asyncio
+import numpy as np
+from concert.base import background, check
+from concert.coroutines.base import broadcast
+from concert.devices.monochromators.base import Monochromator as BaseMonochromator
+from concert.processes.common import ascan
+from concert.coroutines.sinks import Accumulate
+from concert.quantities import q
+
+
+class Monochromator(BaseMonochromator):
+    """
+    Base implementation of a monochromator with the ability to scan a second crystal/multilayer.
+    """
+    def __init__(self, motor_2):
+        """
+        :param motor_2: Motor controlling the tilt of the second crystal or multilayer
+        :type motor_2: concert.devices.motors.base.RotationMotor
+        """
+        self._last_scan_angles = None
+        self._last_scan_counts = None
+        self._motor_2 = motor_2
+        super().__init__()
+
+    @background
+    @check(source='standby', target='standby')
+    async def scan_bragg_angle(self, diode, plot_callback=None, n_points=50, tune_range=0.025*q.deg,
+                               center_point=None):
+        """
+        Scans the second crystal or multilayer. After the scan, the motor is moved back to its
+        initial position.
+
+        A scan can be shown afterwards with :py:func:`.show_tune_scan`. To move the motor to the
+            maximum call :py:func:`.select_maximum` and to go to the center of mass
+            :py:func:`.select_center_of_mass`.
+
+        :param diode: Diode to measure the intensity
+        :type diode: concert.devices.photodiodes.base.Diode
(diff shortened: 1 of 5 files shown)


MarcusZuber:
Open for discussions

tfarago:
What do you want to tube? :smiley:

tfarago:
I don't think it's best idea to put all the showing and stuff to the class. I'd put that elsewhere.

MarcusZuber:
> I don't think it's best idea to put all the showing and stuff to the class. I'd put that elsewhere.

My idea was to provide a way to do this without kowing about ploting the series etc...

tfarago:
> > I don't think it's best idea to put all the showing and stuff to the class. I'd put that elsewhere.
> 
> My idea was to provide a way to do this without kowing about ploting the series etc...

That you can still do for the user, just not in the class because it doesn't belong there IMHO. Why can't we have `concert.processes.[beamline,common].scan_bragg_angle`?

MarcusZuber:
> > > I don't think it's best idea to put all the showing and stuff to the class. I'd put that elsewhere.
> > 
> > 
> > My idea was to provide a way to do this without kowing about ploting the series etc...
> 
> That you can still do for the user, just not in the class because it doesn't belong there IMHO. Why can't we have `concert.processes.[beamline,common].scan_bragg_angle`?

But this will be basically ascan(motor[pos], diode.get_intensity()) ...

tfarago:
You can make a very high-level user convenience function which even plots and does everything. Exactly like you wrote it, just take it out of the class because the scan is actually a higher level functionality which uses the monochromator. I'd like to make the scan "atomic" later, i.e. make sure the shutter is opened and closed and so on. This function can then easily blow up.

MarcusZuber:
Now plot_callback can be plt.show with plt an instance of PyplotViewer.

Question @tfarago, @EliasHamann, @angelicacecilia: What should happen if no plot_consumer is provided? Should I open a new one or should the plotting be ommited?

MarcusZuber:
Maybe, it is nicer to pass the viewer instead of the viewer.show?

EliasHamann:
open for discussion: should we make kind of ascan_brag_angle and dscan_bragg_angle, since the implementation right now always assumes you are already in the center and do the scan from -range/2 to +range/2... (would correspond to dscan_bragg_angle). if you are far off or want to scan arbitrary ranges a functionality with ascan would maybe be nice. (in principle would be ciopy and paste of the existing code just with different start and end points of the scan)...
What do you think?

MarcusZuber:
My ascan is basically a dscan...

What about the possibility to pass a center position? If it is not provided we scan around the current position.

EliasHamann:
this would be possible, however this would not allow asymmetric scans with different ranges on both flanks of the peak; but I don't know if this will be needed. 
maybe we could ask the opinion of the others too...

angelicacecilia:
I find all the solutions very suitable to what we need.
In practice we will never be on the peak of the rocking curve when we put the monochromator in. We will rather be somewhere around and we will need 2-3 scans to get a symmetric scan. In SPEC I usually use the dscan for the first scans, to find the region of interest. For the fine scan I use the ascan and I move the motors absolutely. But as I said at the beginning of my message, I am favourable to whatever working solution make everybody happy.

MarcusZuber:
Now you only provide the `PyplotViewer `instance and you have a `center_point` parameter, where you can specify the center of the scan. If the `center_point` is `None `the current bragg angle is used for the scan.

I will add also the test now.

MarcusZuber:
What about the `show_tune_scan` function? Should it stay there for simple usage of the monochromator or is the user skilled enough to do a plot?

MarcusZuber:
From my side we could merge.

MarcusZuber:
How about to change the `show_tune_scan(self)` to `get_last_tune_scan` that returns just the values?
The the user would just do `plt.show(mono.get_last_tune_scan())` and we don't have the plotting stuff in the monochromator class.

PR #466, opened 2022-02-11
Force Tango to use IPython's loop

MarcusZuber:
Sets the loop for tango devices to the correct ipython-loop


tfarago:
There's nothing more to this, right?

PR #467, opened 2022-02-11
Grating inteferometry experiment

MarcusZuber:
Experiments and addons related to grating inteferometry.

ToDo:

- [x] Test
- [x] Documentation
- [x] Make writing of the results of the Addon more elegant (and correct)
- [x] Fix Tests 

So far I will implement only the stepping. CT-Scans will be added in a separate PR.

### DIFF
--- a/concert/base.py
+++ b/concert/base.py
@@ -1081,6 +1081,7 @@ class Parameterizable(object):
 
         getter_name = 'get_' + param.name
         setter_name = 'set_' + param.name
+        target_getter_name = 'get_target_' + param.name
         if getter_name not in self.__class__.__dict__:
             def get_parameter(instance, wait_on=None):
                 return instance[param.name].get(wait_on=wait_on)
@@ -1089,6 +1090,10 @@ class Parameterizable(object):
             def set_parameter(instance, parameter_value, wait_on=None):
                 return instance[param.name].set(parameter_value, wait_on=wait_on)
             setattr(self.__class__, setter_name, set_parameter)
+        if target_getter_name not in self.__class__.__dict__:
+            def get_target_parameter(instance, wait_on=None):
+                return instance[param.name].get_target(wait_on=wait_on)
+            setattr(self.__class__, target_getter_name, get_target_parameter)
 
         if not hasattr(self, '_set_' + param.name):
             setattr(self, '_set_' + param.name, _setter_not_implemented)
--- a/concert/experiments/addons.py
+++ b/concert/experiments/addons.py
@@ -1,10 +1,12 @@
 """Add-ons for acquisitions are standalone extensions which can be applied to them. They operate on
 the acquired data, e.g. write images to disk, do tomographic reconstruction etc.
 """
+import os
 import logging
+import numpy as np
 from concert.coroutines.base import async_generate
 from concert.coroutines.sinks import Accumulate
-
+from concert.experiments.imaging import GratingInterferometryStepping
 
 LOG = logging.getLogger(__name__)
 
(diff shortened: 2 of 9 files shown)


MarcusZuber:
Mostly depends on https://github.com/ufo-kit/concert/issues/472

MarcusZuber:
With then new walker implementation the tests should pass. I will rebase is when https://github.com/ufo-kit/concert/pull/473 is merged.

MarcusZuber:
New things, that are not directly related to the experiment (but make sense to me):

- I made the `DirectoryWalker.writer` public, so I can manually write files with the same writer-type (like in https://github.com/ufo-kit/concert/blob/73089c0a174a064fbf5bf65936a92ab8a2a4bcc3/concert/experiments/addons.py#L312). 
- Remove Parameter from Parameterizable. This can be used, if a child class should not feature a parameter of its parent class. (here: The grating stepping does not require flatfields and does not have a flatfield-Acquistion, so I delete the num_flats parameter.)

concert/base.py
```diff
@@ -1099,6 +1099,18 @@ def set_parameter(instance, parameter_value, wait_on=None):
         if not hasattr(self, '_get_target_' + param.name):
             setattr(self, '_get_target_' + param.name, _getter_target_not_implemented)
 
+    def _remove_parameter(self, param):
```
tfarago, line 1102: What about lingering `obj.get_parameter` and the same for set. Could that be an issue. The removal should bring the class to the same state as before the parameter installation. This should also be tested.

concert/experiments/addons.py
```diff
@@ -1,16 +1,17 @@
...
+import numpy as np
 from concert.coroutines.base import async_generate
 from concert.coroutines.sinks import Accumulate
 
-
 LOG = logging.getLogger(__name__)
 
 
 class Addon(object):
-
```
tfarago, line 13: trigger-happy pep8 fixer?

tfarago (changes requested):

concert/experiments/addons.py
MarcusZuber: Yes. I have a tool that removes the unnecessary empty lines and I run it over the full file. Should I leave it or put the unrelated ones back?

concert/base.py
MarcusZuber: How about now?

MarcusZuber:
Found something stupid (see https://github.com/ufo-kit/concert/pull/467/commits/ddeed9d141efe1f821a1ce0720f068ee561a11f8). We even missed this in the tests, since we only accessed the Parameter via `device['param'].get_target` and not via `device.get_target_param`. I will also add a Test for this.

MarcusZuber:
@tfarago If you are happy with this now I can squash and merge.

concert/base.py
tfarago: I still find it weird that you need to remove properties. It is an indication that the base class `Radiography` is not general enough. On the other hand, couldn't you just use the number of flats and darks as `num_steps_per_period`, isn't that kind of the number of projections? And you could set it separately for flats and radios...

concert/experiments/addons.py
tfarago: yes please, let's have it clean.

concert/experiments/addons.py
```diff
@@ -203,8 +202,128 @@ def _detach(self):
...
+        if self._experiment.acquisitions[-1] == self._experiment.get_acquisition(
+                "reference_stepping"):
+            await self._compare_and_write()
+
+    async def process_object_stepping(self, producer):
+        self._object_stepping = []
+        async for item in producer:
+            self._object_stepping.append(item)
+        if self._experiment.acquisitions[-1] == self._experiment.get_acquisition("object_stepping"):
+            await self._compare_and_write()
```
tfarago, line 260: The above two methods are almost identical and can have a common parametrized ancestor instead of the repetition.

concert/experiments/addons.py
```diff
@@ -203,8 +202,128 @@ def _detach(self):
...
+        if len(self._reference_stepping):
+            intensity, phase, visibility = await self._process_data(self._reference_stepping,
+                                                                    self._dark_image)
+            self.reference_intensity = intensity
+            self.reference_phase = phase
+            self.reference_visibility = visibility
+
+            await self._write_single_image("reference_intensity.tif", self.reference_intensity)
+            await self._write_single_image("reference_phase.tif", self.reference_phase)
+            await self._write_single_image("reference_visibility.tif", self.reference_visibility)
```
tfarago, line 282: Again a lot of repetition.

concert/experiments/imaging.py
```diff
@@ -127,6 +127,10 @@ def __init__(self, walker, flat_motor, radio_position, flat_position, camera, nu
         super().__init__([darks_acq, flats_acq, radios_acq], walker,
                          separate_scans=separate_scans)
 
+    async def prepare(self):
+        if await self._camera.get_state() != "standby":
+            await self._camera.stop_recording()
```
tfarago, line 132: So useful... We should maybe check all the device states here (but that's a separate story).

concert/experiments/imaging.py
```diff
@@ -857,3 +861,216 @@ async def _take_radios(self):
...
+        :type grating_period: q.um
+        :param stepping_start_position: First stepping position
+        :type stepping_start_position: q.um
+        :param num_periods: Number of periods that are sampled by the stepping
+        :type num_periods: int
+        :param num_steps_per_period: Number of stepping positions per period
+        :type num_steps_per_period: int
+        :param propagation_distance: Distance between the sample and the analyzer grating. Only used
+            by the processing addon to determine the phase shift in angles.
+        :type propagation_distance: q.mm
```
tfarago, line 905: Could you convert these to the ones which are used throughout the rest of the concert? I mean
```
.. py:attribute:: foobar
    descr

```

tfarago: Hmm, but that adds a nicely indented "parameters" section to the docs, so leave it.

concert/experiments/imaging.py
```diff
@@ -857,3 +861,216 @@ async def _take_radios(self):
...
+            await self._camera.stop_recording()
+        await self._camera.set_trigger_source("SOFTWARE")
+        async with self._camera.recording():
+            for i in range(await self.get_num_periods() * await self.get_num_steps_per_period()):
+                await self._stepping_motor.set_position(i * step_size +
+                                                        await self.get_stepping_start_position())
+                try:
+                    await self._camera.trigger()
+                except:
+                    pass
```
tfarago, line 1055: Never ever let an exception slip.

tfarago (changes requested):

concert/experiments/imaging.py
```diff
@@ -857,3 +861,216 @@ async def _take_radios(self):
...
+        This function calls stop_sample_exposure() at the end of the experiment run.
+        """
+        await self.stop_sample_exposure()
+
+    async def start_sample_exposure(self):
+        """
+        This function must implement in a way that the sample is exposed by radiation, like opening
+        a shutter or starting an X-ray tube.
+        """
+        raise NotImplementedError()
```
tfarago, line 1069: In the manager PR you are raising without (), we should keep it consistent throughout concert.

MarcusZuber: OK

concert/experiments/addons.py
MarcusZuber: Do you have a suggestion how to make this in a clean way?

concert/experiments/addons.py
MarcusZuber: Done

concert/base.py
MarcusZuber: How about leaving the `_remove_parameter` in if we need in in the future?

concert/experiments/addons.py
MarcusZuber: I have an idea how to restructure this.

MarcusZuber: Done

tfarago:
You can squash and merge :+1:

tfarago:
Why does it say I request one change but doesn't show it? Did you perhaps force-push so it got lost in the limbo somewhere?

tfarago (approved):

tfarago:
> Why does it say I request one change but doesn't show it? Did you perhaps force-push so it got lost in the limbo somewhere?

When I approved the new review it disappeared. But I didn't request any changes, the review was just pending. Oh man, they still have some way to go with the reviews.

MarcusZuber:
I wouldn't squash more since everithing besides 398a22f44e9ae7cbdef426b22d9c0fa36a19c419 is only partially related to this PR.

tfarago:
:fireworks:

PR #471, opened 2022-02-23
Ascan include last

tfarago:
Allow endpoint of `ascan` and `dscan` to be included and relax unit requirements.


tfarago:
I made `include_last=True` by default because I can imagine that's the behavior one would expect...

MarcusZuber (approved):
I'm fine with it.

PR #473, opened 2022-03-08
Enable locking and unlocking walkers

tfarago:
And copy walkers in acquisitions consumers.

### DIFF
--- a/concert/experiments/addons.py
+++ b/concert/experiments/addons.py
@@ -156,11 +156,9 @@ class ImageWriter(Addon):
         """Wrap the walker and write data."""
         async def wrapped_writer(producer):
             """Returned wrapper."""
-            try:
-                self.walker.descend(acquisition.name)
-                await self.walker.write(producer)
-            finally:
-                self.walker.ascend()
+            async with self.walker:
+                writer = self.walker.create_writer(producer, name=acquisition.name)
+            await writer
 
         return wrapped_writer
 
@@ -182,9 +180,14 @@ class OnlineReconstruction(Addon):
     async def _reconstruct(self, producer):
         await self.manager.backproject(producer)
         if self.walker:
-            with self.walker.inside(self.slice_directory):
+            async with self.walker:
                 producer = async_generate(self.manager.volume)
-                await self.walker.write(producer, dsetname='slice_{:>04}.tif')
+                writer = self.walker.create_writer(
+                    producer,
+                    name=self.slice_directory,
+                    dsetname='slice_{:>04}.tif'
+                )
+            await writer
 
     def _attach(self):
         if self._do_normalization:
--- a/concert/experiments/base.py
+++ b/concert/experiments/base.py
@@ -221,20 +221,21 @@ class Experiment(Parameterizable):
         iteration = await self.get_iteration()
         separate_scans = await self.get_separate_scans()
 
+        if self.walker:
+            if separate_scans:
+                self.walker.descend((await self.get_name_fmt()).format(iteration))
+            if os.path.exists(self.walker.current):
+                # We might have a dummy walker which doesn't create the directory
+                handler = logging.FileHandler(os.path.join(self.walker.current,
+                                                           'experiment.log'))
+                formatter = logging.Formatter('%(asctime)s - %(name)s - %(levelname)s '
+                                              '- %(message)s')
+                handler.setFormatter(formatter)
+                self.log.addHandler(handler)
+        self.log.info(await self.info_table)
+        LOG.debug('Experiment iteration %d start', iteration)
+
         try:
-            if self.walker:
-                if separate_scans:
-                    self.walker.descend((await self.get_name_fmt()).format(iteration))
-                if os.path.exists(self.walker.current):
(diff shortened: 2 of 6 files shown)


MarcusZuber:
How about putting the copy etc. in the `inside` context manager?

You could use it then like:
```
with walker.inside(dataset) as local_walker:
   local_walker.write ....
```

MarcusZuber:
And the inside like

```
@contextmamager
def inside(self, name):
    local_writer = copy(self)
    local_writer.descend(name)
    return(local_writer)
```

tfarago:
No no, I did that in a hurry and there must be a better way. And also if a method creates another instance of the class, that raises a red flag to me.

MarcusZuber:
I don't see the problem to copy a class, that does not hold any data and should be deleted when the `inside` is exited.
Otherwise, we have to tell the user, that he needs to make a copy any time he uses the walker....

MarcusZuber:
To prevent that the user makes to many copies that don't get deleted:
```
@contextmamager
def inside(self, name):
    local_writer = copy(self)
    local_writer.descend(name)
    try:
        yield local_writer
    finally:
        del local_writer
```

And the root could be changed in the copy, so you can not break out of the inside-direcrory.

MarcusZuber (commented):
There is one "use-case", where the logging-setup can cause an exception: When the file system is full.

tfarago:
> There is one "use-case", where the logging-setup can cause an exception: When the file system is full.

OK, but then the experiment _should_ raise an exception before it even starts, that's OK.

tfarago:
Alright, now it should be much better.

tfarago:
@sgasilov, please check if this could break your concert scripts and GUI, I think not, but still.

MarcusZuber:
But now you lock the writer inside of the `inside`. This means a second `inside `call will create an exception.

tfarago:
No, why? The second `inside` will wait until the lock is released and then do its thing.

MarcusZuber:
Will it release the lock? But also when it blocks until the first is done I don't like it so much. E.g. I have the ImageWriter addon and an Addon that writes flat-corrected frames, that both (should) use an `inside`. Then one of them will be blocked for the whole Acquisition.

MarcusZuber:
Ok. The `inside` is only active for creating the write and then leaves the context.

What still bothers me a bit is, that if you directly use the `inside` on multiple locations, the lock will block the code (when the user does not understand that he is only allowed to use it to get the writer inside the `inside`). 

What do you think about changing the `inside` to private and making a `create_writer_inside` function?

tfarago:
Alright, this is how I had it originally but thought needing two "withs" in addons was weird, but actually why not like this, it's also well-aligned with the `async with device` mechanism.

MarcusZuber:
I would propose to move the `wrapper_writer` to the `walker` (with the name as argument).
I use the walker.writer also without the addons and this would make it easier to handle.

The rest I like it now!
Tomorrow morning I will test it and check if I find a way to break it :smirk:.

tfarago:
That's taking it too far. If the lock is acquired in some internal method and the user wants to use it as well it will cause a deadlock:
```python
async with walker:
    walker.wrapped_writer(...) # deadlock
```
This stuff is for "expert" users for very fine control so I think the responsibility should stay on the programmer side (explicit is better than implicit), i.e. they need to think about what they are doing.
We could remove the second context manager by having `create_writer` use the the `inside` if you like.

MarcusZuber:
> We could remove the second context manager by having `create_writer` use the the `inside` if you like.

This I like. The `create_writer` would not lock the `walker` but would always give me a writer in a subfolder. Then the user can call as many `walker.create_writer` as he likes.
In the "normal" experiment the `experiment.run()` would do the `descend`/`ascend` and in the addons the `create_writer` would be used (which does not move the `walker.current` away from the experiment base folder).

Do we really need to make the `walker.inside` public then? For almost all cases the `create_writer` could be used and the rest could be done with the lock and ascend/descend.

MarcusZuber:
Almost.
When you create multiple writers "at the same time" you still have the issue with the messed up directories. I had the idea, that the `create_writer` would not change the `walker.current`. But this requires many changes.

What do you think?

PS: It's a pity that asyncio does not feature reentrant locks. Then it would be easy. Here https://github.com/python/asyncio/issues/439 they decided, that nobody needs this...

tfarago:
> What still bothers me a bit is, that if you directly use the inside on multiple locations, the lock will block the code (when the user does not understand that he is only allowed to use it to get the writer inside the inside).

Now we are starting to spin in circles. If I revert it to how it was before we will get to the cited point above again. I want to keep it like it is now becuase
- it's aligned with the rest of the locking code
- user has full control
- this is advanced stuff, hence users must know what they are doing and for that they need the previous point

MarcusZuber (approved):
I like it.

When I was looking at the code: Why is there `return self` in the ascend/descend? And when there is a reason, why does `home()` not return the walker-instance?

tfarago:
Becuse you can do `walker.descend('foo').descend('bar'); walker.ascend().ascend()`. There is no point for this in `home`.

PR #475, opened 2022-03-17
Upgrade uca net

tfarago:
Needs [uca-net#13](https://github.com/ufo-kit/uca-net/pull/13). I put it in the camera which breaks the same principles as by the libuca implementation I closed. On the other hand, I _do_ use _self_ in the communication and it's more convenient to use as opposed to the case when it would be outside of the class, e.g. `ucad_push_request(num_images, ucad_host, ucad_port)`. I will let it rest for a second, it's minimal so we can always change it.

### DIFF
--- a/concert/devices/cameras/uca.py
+++ b/concert/devices/cameras/uca.py
@@ -1,9 +1,11 @@
 """
 Cameras supported by the libuca library.
 """
+import asyncio
 import functools
 import logging
 import numpy as np
+import struct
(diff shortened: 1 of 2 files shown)


MarcusZuber:
This https://github.com/ufo-kit/concert/pull/409 should also go here then, because it is only relevant for uca-net cameras.

tfarago:
This will be handled as a part of #481

PR #478, opened 2022-03-22
Hercules

tfarago:
Various improvements which were needed for my demonstration of Concert during Hercules.

### DIFF
--- a/concert/coroutines/sinks.py
+++ b/concert/coroutines/sinks.py
@@ -1,5 +1,6 @@
 import asyncio
 import logging
+import numpy as np
 from concert.config import AIODEBUG
 from concert.coroutines.base import background
 
@@ -44,13 +45,16 @@ async def null(producer):
 
 
 class Accumulate(object):
-    """Accumulate items in a list or a numpy array if *shape* is given, *dtype* is the data type.
+    """Accumulate items in a list or a numpy array if *shape* is given, *dtype* is the data type. If
+    *reset_on_call* is True, the saved values will be overwritten every time the accumulator is
+    called, otherwise they will be appended.
     """
 
-    def __init__(self, shape=None, dtype=None):
-        import numpy as np
-
-        self.items = [] if shape is None else np.empty(shape, dtype=dtype)
+    def __init__(self, shape=None, dtype=None, reset_on_call=True):
+        self._shape = shape
+        self._dtype = dtype
+        self.reset_on_call = reset_on_call
+        self.items = [] if shape is None else np.empty((0,) + self._shape[1:], dtype=self._dtype)
 
     @background
     def __call__(self, producer):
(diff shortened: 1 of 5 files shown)


concert/coroutines/sinks.py
```diff
@@ -59,23 +63,36 @@ def __call__(self, producer):
...
         if isinstance(self.items, list):
             return self._process(producer)
         else:
             return self._process_numpy(producer)
 
+    def reset(self):
+        if isinstance(self.items, list):
+            del self.items[:]
+        else:
+            self.items = np.empty((0,) + self._shape[1:], dtype=self._dtype)
```
MarcusZuber, line 78: `self.items = np.empty_like(self.items)` should do same, or?

concert/ext/ufo.py
```diff
@@ -682,27 +682,19 @@ def score(vector):
...
+            self._num_received_projections += 1
+            async with self._producer_condition:
+                self._producer_condition.notify_all()
 
-            if self._num_received_projections < self.args.number:
-                await run_in_executor(copy_projection, projection)
-                self._num_received_projections += 1
-                async with self._producer_condition:
-                    self._producer_condition.notify_all()
+    async def _copy_projections(self, producer):
```
MarcusZuber, line 695: I find the naming a bit confusing since `_copy_projections` calls `_process_projections` which internally calls `copy_projections`.

concert/coroutines/sinks.py
tfarago: No, self.items can grow and you want to reset to the shape given in contructor.

concert/processes/common.py
```diff
@@ -216,20 +216,20 @@ async def acquire_frames_360(camera, rotation_motor, num_frames, shutter=None, f
          position_eps=Numeric(1, q.deg), max_iterations=Numeric(1),
          initial_x_coeff=Numeric(1, q.dimensionless), initial_z_coeff=Numeric(1, q.dimensionless),
          shutter=Shutter, flat_motor=LinearMotor, flat_position=Numeric(1, q.m),
-         y_0=Numeric(1), y_1=Numeric(1), get_ellipse_points_kwargs=None, frame_consumer=None)
+         y_0=Numeric(1), y_1=Numeric(1), get_ellipse_points_kwargs=None, frame_consumers=None)
```
MarcusZuber, line 219: Just for curiosity: What was your use case here for multiple frame_consumers? Why not just using a broadcast coro?

concert/ext/ufo.py
tfarago: Yeah, but it calls `copy_projection` defined inside the method and nothing of this is part of public API anyway.

concert/coroutines/sinks.py
MarcusZuber: self.items = np.empty_like(self.items[0]) ?

concert/processes/common.py
tfarago: I was storing in an accumulator and sending to live view at the same time. Then I showed `viewer.show(np.mean(acc.items[10:20]))` to show how the needle travelled (360 deg in 10 projections).

PR #479, opened 2022-03-23
Enable running the loop from within a running loop

tfarago:
Which allows us to execute async code in a loop explicitly even when we
are an async code. It is now possible to run device.param = 'foo' from
both, non-async code and async-code. In short, this is possible:

## Version 1
```python
class MyDevice(Device):
    foo = Parameter()

    def __init__(self, value=1):
        super().__init__()
        self._foo = None
        self.foo = value

    async def _set_foo(self, value):
        self._foo = value

    async def _get_foo(self):
        return self._foo


async def factory():
    # This would have failed before with a RuntimeError: loop already running
    return MyDevice(value='from factory')

md = MyDevice()
```

## Alternative: version 2
```python
class MyDevice(Device):
    foo = Parameter()

    def __init__(self):
        super().__init__()
        self._foo = None

    @classmethod
    async def create(cls, value=1):
        self = MyDevice()
        await self.set_foo(value)

        return self

    async def _set_foo(self, value):
        self._foo = value

    async def _get_foo(self):
        return self._foo

async def factory():
    # This would have failed before with a RuntimeError: loop already running
    return await MyDevice.create(value='from factory')

md = run_in_loop(MyDevice.create())
```
## Alternative: version 3

We can add `__ainit__` to `Parameterizable`. If the child class implements `__ainit__`, it will be `RuntimeError` to try to construct the object directly via `MyClass()`, one will be *forced* to use `await MyDevice.ainit()`.

```python
class MyDevice(Device):
    foo = Parameter()

    def __init__(self):
        super().__init__()
        self._foo = None

    async def __ainit__(self, value=1):
        await self.set_foo(value)

    async def _set_foo(self, value):
        self._foo = value

    async def _get_foo(self):
        return self._foo

md = MyDevice() # RuntimeError
md = run_in_loop(MyDevice.ainit(value='from session'))
```

If we want this, we *must*
- [ ] check if [nest-asyncio](https://pypi.org/project/nest-asyncio/) wouldn't be a better option, especially because real devices could tie themselves to the event loop in the main thread (which tango does) and that would need to be handled for all device types (for tango it is inside ankaconcert)
- [ ] check thoroughly with real devices (tango, ...)
- [ ] test this _truly heavily_ inside concert
- [ ] make sure that all relevant occasions (especially in `concert.base`) of `run_in_loop` are replaced by this

### DIFF
--- a/bin/concert
+++ b/bin/concert
@@ -1,9 +1,9 @@
 #! python
 
+import ast
 import asyncio
 import argparse
 import contextlib
-import functools
 import logging
 import os
 import re
@@ -12,7 +12,6 @@ import shutil
 import signal
 import subprocess
 import tempfile
-import traceback
 import zipfile
 import concert
 import concert.config
@@ -39,9 +38,11 @@ def cmp_versions(v1, v2):
     return (n1 > n2) - (n1 < n2)
 
 
-def get_prompt_config(session_name):
-    """Customize the prompt for the *session_name*."""
+def get_prompt_config(path):
+    """Customize the prompt for session in *path*."""
     from IPython.terminal.prompts import Prompts, Token
+
+    session_name = os.path.splitext(os.path.basename(path))[0]
 
     class MyPrompt(Prompts):
         def in_prompt_tokens(self, cli=None):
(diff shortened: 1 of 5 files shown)


MarcusZuber:
Looks promising. I will test it on Friday with tango, UCA and UFO and the viewers.
Do we have anything else external, that could interfere with the loops?

MarcusZuber:
> Looks promising. I will test it on Friday with tango, UCA and UFO and the viewers. Do we have anything else external, that could interfere with the loops?

Unfortunately, I have to postpone the tests to next Wendsday...

MarcusZuber:
The version 2 is basically a manual version of this https://pypi.org/project/asyncinit/ .

tfarago:
Yes, but we cannot have async constructors, otherwise you would get syntax errors on concert start session.

MarcusZuber:
> Yes, but we cannot have async constructors, otherwise you would get syntax errors on concert start session.

But when I need the async `create()` function (option 2) the same error would appear.

tfarago:
Yes, that's right.

MarcusZuber:
Did you already test nest-asyncio? I  just had a short look at the module and I'm not sure if this would not create more problems than solving...

tfarago:
I didn't think too deeply about the consequences but there must be a good reason why the loops are not re-entrant by default and of course the library has to do ugly things to patch it. Do you see some concrete problems with it?
In general, I would much rather create a policy on what to use when and then have the developers stick to it which is both safer and explicit. I just don't know yet how to enforce it, meaning "forbid" trying to nest coroutines.

MarcusZuber:
What I tested so far:

I used (many) tango-motors (with green-mode = asyncio), an uca-net camera and a CT-scan (based on `concert.experiments.XrayTube.SteppedTomography`) + OnlineReco.
In the experiment `__init__` some properties are set. Creating an experiment instance worked "by-handy" and via a async factory method.

The setting of the properties in the `__init__` put this in the log "Loop already running, expediting QuantityValue._set to a thread", what was expected.

The experiment as well as the direct access to the devices worked.

tfarago:
That's good, I still don't like this solution :smile: Is this urgently needed?

tfarago:
I just added "Alternative 3" which I like because it makes things explicit.

MarcusZuber:
I thought of something simmilar. But the user/dev needs to make sure, that the ainit is called. It would be nice to add a `run_in_loop(ainit())` in the `__init__`, but this will work only with complicated *kwargs parameters.

tfarago:
That's ensured becuase they will get this error when trying to use the normal constructor

```python
RuntimeError: MyDevice must be initialized in async code with MyDevice.ainit(...)
```

So they won't get their object and we tell them what they need to do. Calling `run_in_loop(ainit())` in `__init__` sends us to square one. You will get loop already running exceptions if used from within an async code.

MarcusZuber:
Here some thoughts I had about the different options:

Option 1: Most clean from the "outside". One would just use the normal constructor and could access all the properties as they would be `@property`. Inside the class `run_in_loop_or_thread` could maybe cause some troubles.

Option 2: Dangerous that only the constructor is called, but the async-stuff is not executed. Maybe changing the default_state to "not initialized" would be something to add then.

Option 3: Simmilar to 2 but prevents the direct call of the contrsuctor.

If we coose 2 or 3 I would suggest to make this consitent to all objects in concert (I already can hear the voice of some people compaining, why some shings are different in some devices). What I don't like, that these implementations are different from normal python classes, what can be quite confusing to python newcomers. If we would change all contstructors/ainit s in concert the async-init would maybe also something to try...

tfarago:
We are in for some very serious trouble if we proceed with this. E.g. events, locks and who knows what else is tied to concrete loops, so for example `async with walker` will not work, `async with parameter` will not work, etc.

tfarago:
I am looking into the async constructor option, if I manage to allow `await` syntax in concert sessions it would be a way to go, let's see.

MarcusZuber:
Looks promising. I will play with it on Friday and see if I can break it 😜

tfarago:
Closing this in favor of #482 with a cleaner version of "Alternative 3".

PR #480, opened 2022-04-04
First qt-widget-wrapper implementation

MarcusZuber:
First attempt to write a qt-widget-wrapper that can be applied to a `Parameterizable`.

The code is still quite chaotic but in the first tests it did what it should be.

An example session:
```
import logging
from inspect import iscoroutinefunction
import concert
concert.require("0.30.1")

from concert.session.utils import ddoc, dstate, pdoc, code_of

from concert.gui.parameterizable import ParameterizableWidget

lin_motor = LinearMotor()
lin_motor_widget = ParameterizableWidget(lin_motor)
```
Then `lin_motor_widget.show()` opens the widget. Can be also used as a general widget in qt layouts etc.

TODOs / Questions:
- [] How to handle `Properties` with arbitrary data types? Maybe add a string-to-property function to the property?
- [] How to add functions? Create subclasses for each device?
- [] Test performance of (ugly manual) event-loop.


MarcusZuber:
With the Widgets one can build a standalone application like

```python
from PyQt5.QtGui import QApplication
from concert.devices.samplechangers.dummy import SampleChanger
from concert.gui.parameterizable import ParameterizableWidget

if __name__ == '__main__':
   app = QApplication([]) # or args etc.
   changer = SampleChanger()
   changer_widget = ParameterizableWidget(changer)
   changer_widget.show() # or an arbitrary combination in layouts etc.
   app.exec()
```

I also wanted to have the possibility to open a Widget from a running concert session. Therefore, I added this manual "qt event loop". The possibility to open a widget in a running session is required, if devices do not allow access from multiple connections (or when parameters/state is handled in concert -> that's why I don't like the `transition` decorator).
Also for checking within a running session if a device is working correctly (without setting up a special session) is I feature that I would like to have. (With some physical devices I have serious trust issues :wink: and to have a small window where state/parameters are polled from time to time can help a lot)

So far I wrapped all async slots that qt sees only non-asyncio blocking functions. I didn't test qt-asyncio compatible event loops like asyncqt or quamash with tango etc. But changing the event loop within a running concert session (not a standalone python program) does not work.

tfarago:
Before we continue here this will require a discussion with the others, we could try to chat in the week after Easter.

MarcusZuber:
Sure. I just wanted to dump everything from my mind here, otherwise I can not remember until the next discussions :wink: 

I just played with it in the lab and it feels quite nice and responsive.

MarcusZuber:
Note to myself: Replace the slot-wrapper with the qasyncio wrapper to make it more compatible with standalone qasyncio applications.

tfarago:
When I exit `python bin/gui_standalone.py` I get a segmentation fault :fearful:

concert/gui/parameterizable.py
```diff
@@ -0,0 +1,227 @@
...
+        super().__init__()
+        self._param = param
+        self.name_label = QLabel(param.name)
+        self.read_button = QPushButton('read')
+        self.write_button = QPushButton('write')
+
+        self.read_button.clicked.connect(self.read)
+        self.write_button.clicked.connect(self.write)
+
+        # TODO: in standalone the run_in_loop is not working -> fix
```
tfarago, line 122: This should be history if we manage to have the widget like this:
```python
class ParameterWidget(QWidget, AsyncObject):
   def __ainit__(self, param):
        super().__init__()
        await super().__ainit__()
```

MarcusZuber:
> When I exit `python bin/gui_standalone.py` I get a segmentation fault :fearful: 

I'm quite satisfied in this state, that I only get this when exiting...

I already think to know the cause. When the full-async stuff is working I will fix it.

MarcusZuber:
So far I'm quite happy with the functionality and interface. We should discuss how to proceed.

PR #482, opened 2022-04-29
Asyncify the complete codebase

tfarago:
Replace `obj = Parameterizable()` with `obj = await Parameterizable()` and make sure `await` is allowed in session files in global scope. Supercedes #479 in favor of having everything cooperate instead of having to deal with various loops and maybe even threads.

### Motivation
We have a loop which exists throughout the life of a concert session and often devices (like in the case of Tango), events and locks are bound to it. When concert device classes are instantiated they often need to set some properties on the devices (e.g. by a Tango proxy), i.e. they need to invoke async code, which is not possible in `__init__`. Hence, we have two options:
- run async code inside another loop _and thread_ like in #479, which leads to all kinds of pitfalls and headaches
- enable "natively" async code in `__init__`, which is what we do in this PR

### Features
This PR enables us to use `await` in the session file (outside of a function) and adds an awaitable constructor `__ainit__`, so one can do `obj = await Class(...)`. This way, device constructors can use `await` in order to communicate with the devices already in the constructors.

Regarding things like `motor.position = 1 * q.mm` called from within an async code, this will stay a `RuntimeError` becuase with this PR it will be possible to use getters and setters properly, e.g. call `await motor.set_position(1 * q.mm)` in an `__ainit__(...)` of the motor class.

### Notes
- Decorator-based [asyncinit](https://github.com/kchmck/pyasyncinit) causes `__init__` to be called twice if people accidentally decorate both parent and child classes.
- Having a custom `async def __new__(cls, *args, **kwargs)` which calls both `__new__` and `__ainit__` would prevent the `__init__` to be called because it doesn't return an instance of a class but a coroutine, in which case Python won't call `__init__`. So technically, this would work in simple scenarios, however,  it can cause trouble in cooperative inheritance, where a non-async code could call an async sibling via `super().__new__(...)` and get a coroutine instead of an object. This means that the non-async classes would stop working when used in mixed inheritance trees.
- A metaclass enables us to go veeery deep into the object construction process and allows us to separate *creation* and *initialization*. Thus, `__new__` can stay as it was and we "only" replace the initialization part. (A nice post on object creation can be found [here](https://eli.thegreenplace.net/2012/04/16/python-object-creation-sequence), it's from 2012 and the C code changed a little but the most important function calls stayed)
- Adding `async def __anew__(cls, *args, **kwargs)` is not needed at the moment, at least I have a hard time justifying why the object _construction_ should need any async code.
- `Parameterizable` will inherit from `AsyncObject`, the reason is twofold:
  - parameters are there to be used asynchronously and it is very well probable that when a `Parameterizable` makes its parameters, it will also want to use them in an async-way already in its constructor
  - It would cause a lot of confusion about needing to instantiate some parameterizables with `await Foo(...)` and others with `Bar(...)`, so the code usage is unified by this

### Conversion helpers

- Replace `def __init__(...)` with `async def __ainit__(...)` for all files in DIR
  - `find $DIR -type f -exec sed -i 's/def __init__/async def __ainit__/g' '{}' \;`
- Replace `super(...).__init__(...)` with `await super(...).__ainit__(...)` for all files in DIR
  - `find $DIR -type f -exec sed -i 's/super\(.*\).__init/await super\1.__ainit/g' '{}' \;`
- When there are white spaces in front:
  - `sed -i 's/\([[:space:]]*\)\(.*\)\.upper = \(.*\)/\1await \2.set_upper(\3)/g' concert/tests/unit/test_parameter.py`

### Todo

- [x] make `Parameterizable` inherit from `AsyncObject` and thus change most of the codebase to have `__ainit__` constructor
- [x] get rid of `run_in_loop_thread_blocking`
- [x] README.md
- [x] tests
- [x] todos in code for temporary fixes and workarounds
- [x] docs
- [x] Check if we can change `ip_config.InteractiveShellApp.exec_lines` for something like `file_to_run` or similar in the light of being able to import modules with `await` top-level statements (no we cannot, stays like it was for now)
- [x] flake8 + with this PR's unrelated issues commited last before release
- [x] concert-examples
- [x] Prepare new release because of the broken API


MarcusZuber:
Ok, first "issue":

The `mydevice await = Device()` works only in session files. We usually split the instantiation of devices and sessions files like

File with dev-instances
devices_of_beamline_part_X.py:
```
mydevice = (await) Device()
```

And then the session:
```
from devices_of_beamline_part_X import mydevice
```

How can I get such a behavior to work?

MarcusZuber:
The rest seams (when just playing in an unstructured way with it) to work.

MarcusZuber:
When I was testing, I realized, that be should try to make everything a child of AsyncObject (or Parameterizable).
I was a bit confused, that I had to write this:
```
from concert.storage import DirectoryWalker
from concert.ext.viewers.PyplotImageviewer

viewer = await PyplotImageViewer()
walker = DirectoryWalker()
```

concert/base.py
```diff
@@ -66,12 +81,25 @@ def _find_in_dict(dictionary):
...
                 break
     finally:
         # Cleanup as python docs suggest
         del frames
 
+    if not instance_name:
+        # Try coroutines
+        tasks = asyncio.all_tasks(loop=asyncio.get_running_loop())
+        for task in tasks:
+            # Python 3.7 compatibility, later task.get_coro()
```
MarcusZuber, line 94: I thought we are requiring >=3.8

MarcusZuber (approved):
WOW!

concert/base.py
tfarago: Strictly speaking yes, but only for tests, for the codebase I stick with 3.7

MarcusZuber (approved):

bin/concert
```diff
@@ -432,69 +445,33 @@ class StartCommand(SubCommand):
...
+                                                                 '_ipython_setup.py')]
+        ip_config.TerminalInteractiveShell.prompts_class = get_prompt_config(path or 'concert')
+        ip_config.TerminalInteractiveShell.loop_runner = 'asyncio'
+        ip_config.TerminalInteractiveShell.autoawait = True
+        ip_config.InteractiveShell.confirm_exit = False
+        ip_config.TerminalIPythonApp.display_banner = False
+        ip_config.Completer.use_jedi = False
+
+        # Now we start ipython with our configuration
+        IPython.start_ipython(argv=[], config=ip_config)
```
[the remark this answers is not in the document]
tfarago: Thanks, I also don't know why it was like that before, maybe this didn't exist back then.

MarcusZuber:
I found one small problem. (But I'm afraid the fix could be complicated...)

You can no longer see the constrcutor parameteris in the print of the docstrings of the constructor since the `__init__ `takes only args and kwargs.

```python
tomography > from concert.directors.scanning import XYScan                                                                              

tomography > XYScan?                                                                                                                    
Init signature: XYScan(*args, **kwargs)
Docstring:      Director to scan a specimen within a plane.
File:           /usr/local/lib/python3.8/dist-packages/concert-0.30.1.dev0-py3.8.egg/concert/directors/scanning.py
Type:           AsyncType
Subclasses:   
```

tfarago:
> You can no longer see the constrcutor parameteris in the print of the docstrings of the constructor since the __init__ takes only args and kwargs.

If we teach people to use `help(XYSscan)` or even `XYScan??` instead of `XYScan?` they will see the `__ainit__`, so I don't see this as an issue.

> Did you test if the auto-docs are build properly?

Of course not and of couse it didn't build correctly. but should be fixed now.

MarcusZuber:
About b49ef839b5cffb41a4911768d344d6765496d3f7 :

How about overwriting the __init__ signature in the AsyncObject class?

Maybe something similar to:

```python
class AsyncObject:
   __init__.__signature__ = inspect.Signature(AsyncObject.__ainit__)

```

(Hm, it should be the signature of the final version...)

With such an implementation I would be confident, that also Jedi etc. would understand this.

(I'm only on my phone today and I can not test stuff. I'm just dumping my ideas here.)

MarcusZuber (approved):
I didn't check all the details, but tested it "in operation". So far I was not able to break it :+1: 

From my side we can merge it!

MarcusZuber:
You will not like this.

When devices (or AsyncObjects) are created in a session-file and in an imported module you get:

StartError: Module `async_import' uses `await' outside of `async def' functions but at the same time tries to execute coroutines in a loop which is not allowed, please remove the loop-running code from the module and try again


Example:
async_session.py
```python
"""# *async-test* session manual

## Usage

## Notes
"""

import logging
from inspect import iscoroutinefunction
import concert
concert.require("0.32.0")

from concert.session.utils import ddoc, dstate, pdoc, code_of

l = await LinearMotor()

from async_import import dev
```

async_import.py
```python
from concert.devices.motors.dummy import RotationMotor

dev = await RotationMotor()
```

If all AsyncObject instances are created within the session it works and also if all are in the imported module it works. I they are mixed it crashes.

PR #484, opened 2022-05-02
add stop to experiment

MarcusZuber:
Add a possibility to stop a running experiment without killing all tasks.


concert/experiments/imaging.py
```diff
@@ -539,7 +539,8 @@ async def _finish_radios(self):
         """
         if self._finished:
             return
-        await self._tomography_motor.stop()
+        if await self._tomography_motor.get_state() == "moving":
```
tfarago, line 542: Same for ContinuousSpiralTomography

tfarago:
How is this different from calling `ex.run().cancel()`.

MarcusZuber:
It's class method, so you can stop even when you forget to store the handle to exp.run()  and it is more structurized stop that tries to finish stuff in a that you can actually re-run the experiment afterwards (e.g. if a cont. motor is moving the cancel would only stop it in the acceleration phase).

tfarago:
I see, alright.

MarcusZuber:
I thought to catch the CancelError and then make a structured stop, but then this would shadow the cancel of the motors etc. which might be also required in some cases.

concert/experiments/imaging.py
```diff
@@ -279,7 +284,13 @@ async def _produce_frames(self, number, **kwargs):
         """
         await self._camera.set_trigger_source("AUTO")
         async with self._camera.recording():
+            if isinstance(self, Tomography):
+                self.log.info("First frame started at position %s" %
+                              await self._tomography_motor.get_position())
+
             for i in range(int(number)):
+                if self._stop:
```
tfarago, line 292: Also, instead of checking this in _every_ experiment class, we can patch catch a stop request in the `Acquistion`, namely you can wrap the `broadcast` into something which accepts a stop request.

MarcusZuber: But what will it change? I have to write everywhere, where right now is `if self._stop` a handler of the stop request?

concert/experiments/base.py
```diff
@@ -318,6 +321,22 @@ async def _run(self):
...
+        as possible ctrl-k should be used.
+        This function blocks until the experiment is completely stopped.
+        """
+        self._stop = True
+        LOG.info("Experiment stopped.")
+
+        async def condition():
+            return await self.get_state() != "running"
+
+        await wait_until(condition)
```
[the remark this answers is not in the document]
MarcusZuber: Ah, it was not there yet, when I wrote this.

concert/experiments/imaging.py
MarcusZuber: But a clean stop is very experiment specific. So I abort all the acquisitions and then call a clean-up function, that has to take care that everything is back in a well defined state?

tfarago: In all your producers you just do `if stop then break`, so no clean up. And of course if special handler is needed you can always override the `Experiment.stop` method.

MarcusZuber:
How about this? It is basically the cancel, but with a different exception. So in a generator I can catch it and make a structured stop.

MarcusZuber:
Oh no:

[asyncio.Task](https://docs.python.org/3/library/asyncio-task.html#asyncio.Task) inherits from [Future](https://docs.python.org/3/library/asyncio-future.html#asyncio.Future) all of its APIs except [Future.set_result()](https://docs.python.org/3/library/asyncio-future.html#asyncio.Future.set_result) and [Future.set_exception()](https://docs.python.org/3/library/asyncio-future.html#asyncio.Future.set_exception).

MarcusZuber:
Super ugly, but does what it should do.

PR #486, opened 2022-05-11
Test of gh-actions

MarcusZuber:
Todo:

- [x] Add codecov
- [x] Test on multiple versions of python


tfarago:
Pretty much the same deal as with circleci, do you have compelling arguments? One for circleci is the 6000 min.

MarcusZuber:
As you prefer. I just wanted to try it.

tfarago:
Actions are actually free for public repos, so gonna switch to this because of unlimited build time.

MarcusZuber:
Note: python3.10-test requires https://github.com/ufo-kit/concert/pull/490

tfarago:
Why do you have a separate coverage.yml now?

MarcusZuber:
Then it was easier to run coverage only for one python version.

tfarago:
So it would call it as many times as you have python versions?

MarcusZuber:
The current test.yml iterates over this list of python versions. For sure it would be possible to tell it to do the coverage only once, but splitting it was more simple.

tfarago:
I think the if statement is more simple than a whole new file, plus you use more minutes because you test one more time then necessary, all you need it `if: matrix.python-version == 3.8` in the coverage setup, check [this](https://github.com/ufo-kit/syris/blob/ci/.github/workflows/tests.yml#L36-L44) out.

MarcusZuber:
OK, codecov was more clever than I expected.

From my side we can merge.

The python3.10 I removed until we migrate from nose. My fix depends on the import order, which is different for coverage, nosetest and nose.main() :roll_eyes:

.github/workflows/tests.yml
```diff
@@ -0,0 +1,43 @@
...
+permissions:
+  contents: read
+
+jobs:
+  build:
+
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        python-version: [3.8, 3.9]
```
tfarago, line 18: Why drop 3.10?

tfarago:
You forgot to change the badge in README

.github/workflows/tests.yml
MarcusZuber: The python3.10 I removed until we migrate from nose. My fix depends on the import order, which is different for coverage, nosetest and nose.main() 🙄

MarcusZuber:
I totally forgot the badge. I will fix this tomorrow.

PR #487, opened 2022-05-14
fix missing ascend and try around experiment execution in director

MarcusZuber:
A test, if the walker is at the same position after the execution would be maybe something to check this also in the future...

### DIFF
--- a/concert/directors/scanning.py
+++ b/concert/directors/scanning.py
@@ -1,7 +1,7 @@
 import asyncio
 from concert.quantities import q
 from concert.helpers import arange
-from concert.base import Quantity, Parameter
+from concert.base import Quantity, Parameter, check
 from concert.directors.base import Director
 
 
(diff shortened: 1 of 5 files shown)


MarcusZuber:
What I wrote recently was not so ideal, but now everithing should work and tested...

concert/directors/scanning.py
```diff
@@ -9,12 +9,12 @@ class XYScan(Director):
...
-    x_step = Quantity(q.mm, check=['standby', 'error'])
-    y_min = Quantity(q.mm, check=['standby', 'error'])
-    y_max = Quantity(q.mm, check=['standby', 'error'])
-    y_step = Quantity(q.mm, check=['standby', 'error'])
+    x_min = Quantity(q.mm)
+    x_max = Quantity(q.mm)
+    x_step = Quantity(q.mm)
+    y_min = Quantity(q.mm)
+    y_max = Quantity(q.mm)
+    y_step = Quantity(q.mm)
```
tfarago, line 17: Why are the checks gone?

MarcusZuber: I didn't check in detail, but I had the impression, that the check parameter requires a function and not a list of states. I will have a second look.

tfarago:
:+1:

PR #489, opened 2022-05-16
remove run_in_executor in concert.devices.cameras.uca._grab_real

MarcusZuber:
Without this, the `run_in_executor(self.uca.grab, data)` runs forever. The uca-mock worked without the fix. :confused:

### DIFF
--- a/concert/devices/cameras/uca.py
+++ b/concert/devices/cameras/uca.py
@@ -210,9 +210,9 @@ class Camera(base.Camera):
         data = array.__array_interface__['data'][0]
 
         if index is None:
-            await run_in_executor(self.uca.grab, data)
+            self.uca.grab(data)
         else:
-            await run_in_executor(self.uca.readout, data, index)
+            self.uca.readout(data, index)
 
         return array
 


tfarago:
I don't get what the problem actually is. Without the executor it will block all tasks so that's not a good idea. On the top of that, if you press ctrl-c while in grab your session can hang.

MarcusZuber:
libuca does in the grab some GIL and mutex stuff (https://github.com/ufo-kit/libuca/blob/3a78e6a5a2b83cbe9a1249d9469f6183a036e954/src/uca-camera.c#L1182). Could this interfere with asyncio?

tfarago:
No no, it's not that, it's definitely the GIL. And you see this only in combination with "net", right? Just build libuca on the server end without the python support and all should work as expected. The python stuff is currently off, I will look into it.

tfarago:
Anyway, on net it hangs even without the `run_in_executor`. Is by chance your detector doing something with the gevent loop?

tfarago:
This is known and an easy fix, just use `meson configure -Dwith_python_multithreading=true` on the servers side where `ucad` runs. Although I will look into this so that *nothing* special is required from the user perspective.

MarcusZuber:
Ok. I will test some things with this tomorrow.
Setting with_python_multithreading to false is required to get it to work with python? Seams not so intuitive...

tfarago:
No, it is required to make it work *without* python, which is the case on the server side where you call `ucad camtype`. No Python there...

tfarago:
The problem seems to be in how `ucad` handles things, there is a thread created where the actual camera calls live and that's not so healthy for python, anyway, still looking into this.

MarcusZuber:
So if I run my plugin in the same machine as concert I should compile it with_python_multithreading=true?
I also had a deeper look into my uca-plugin. The only thing I found are std::mutex, but I don't think they can interfere with asyncio.

tfarago:
> So if I run my plugin in the same machine as concert I should compile it with_python_multithreading=true?

try both and let's see what happens.

About the mutex, I am not sure what exactly is it doing so...

MarcusZuber:
Will be solved by https://github.com/ufo-kit/concert/issues/492 .
Until then I will use it the version without the executor for my plugin.

PR #490, opened 2022-05-17
fix nose for python3.10

MarcusZuber:
Somewhere between python3.9 and 3.10 nose does not work anymore.

Nose also stopped development in 2015 and we should think of moving to nose2 or another test framework.

This fixes the current issue of the old nose implementation for now.


MarcusZuber:
For the long run we should discuss there https://github.com/ufo-kit/concert/issues/491 how to proceed.

PR #493, opened 2022-05-20
Generalize unit of flatfield_position and radio_position of imaging experiments

MarcusZuber:
TODO:

- [x] Docs
- [x] Discuss if something more elegant can be done instead of the LaminographyMixin

In `concert.experiments.imaging.Radiography` the `flat_position` and `radio_position` `Quantity` is created with the unit of the `flatfield_motor['position']` property. This allows to use `LinearMotor`s and `RotationMotor`s

### DIFF
--- a/concert/experiments/imaging.py
+++ b/concert/experiments/imaging.py
@@ -86,24 +86,17 @@ class Radiography(Experiment):
     num_projections = Parameter()
     """Number of projection images."""
 
-    radio_position = Quantity(q.mm)
-    """Position of the flat_motor to acquire projection images of the sample."""
-
-    flat_position = Quantity(q.mm)
-    """Position of the flat_motor to acquire images without the sample."""
-
     async def __ainit__(self, walker, flat_motor, radio_position, flat_position, camera, num_flats,
                         num_darks, num_projections, separate_scans=True):
         """
         :param walker: Walker for storing experiment data.
         :type walker: concert.storage.Walker
-        :param flat_motor: LinearMotor for moving sample in and out of the beam.
-        :type flat_motor: concert.devices.motors.base.LinearMotor
+        :param flat_motor: Motor for moving sample in and out of the beam. Must feature a
+            'position' property.
         :param radio_position: Position of *flat_motor* that the sample is positioned in the beam.
-        :type radio_position: q.mm
+            Unit must be the same as flat_motor['position'].
         :param flat_position: Position of *flat_motor* that the sample is positioned out of the
-            beam.
-        :type flat_position: q.mm
+            beam. Unit must be the same as flat_motor['position'].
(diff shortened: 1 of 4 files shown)


tfarago:
We can also take it further (needs to be checked for constructor compatibility though):
```python
async def make_laminography(
    base_cls, walker, flat_motor, tomography_motor, radio_position, flat_position,
    camera, shutter, num_flats=200, num_darks=200, num_projections=3000,
    angular_range=180 * q.deg, start_angle=0 * q.deg, separate_scans=True
):
    ex = await base_cls(
        walker, flat_motor, tomography_motor, radio_position, flat_position,
        camera, shutter, num_flats=200, num_darks=200, num_projections=3000,
        angular_range=180 * q.deg, start_angle=0 * q.deg, separate_scans=True
    )
    ex.install_parameters({"flat_position": Quantity(q.deg), "radio_position": Quantity(q.deg)})
    await ex.set_flat_position(flat_position)
    await ex.set_radio_position(radio_position)

    return ex
```

tfarago:
And also the name is too restricting, I can imagine a normal tomography to have a flat motor which rotates (for whatever weird reason...). So how about `make_tomography_rotational_flat_motor` or something similar?
If we need this right now we can stop here but in the future I can imagine having a real builder class which would then take care of all the construction peculiarities.

MarcusZuber:
What if I read the unit of the flat motor in the Radiography class and set the positions units accordingly? I have the impression this would be the cleanest (from the outside).

tfarago:
I like this idea.

MarcusZuber:
The last commit basically solves the whole thing.
What do you think? If you like it, I will adapt the tests/docs and drop all the un-required stuff.

tfarago:
Perfect. Just don't forget to remove the commented old positions.

concert/experiments/imaging.py
```diff
@@ -1027,13 +1034,15 @@ async def _get_num_flats(self):
...
 
     async def _set_num_flats(self, n):
-        raise AccessorNotImplementedError
+        if n != 0:
+            raise AccessorNotImplementedError
 
     async def _get_num_projections(self):
         return 1
 
     async def _set_num_projections(self, n):
```
MarcusZuber, line 1043: I know this is a bit ugly. I will do some generalizations on the grating experiments soon, then this will look better again.

MarcusZuber:
I want to add checks for the state for all parameters as well. Otherwise you can break to many things, when someone changes them while the experiment is running.

concert/experiments/imaging.py
tfarago: Alright

tfarago (approved):
Even flake8 is happy :+1:

PR #497, opened 2022-05-27
Aimport future statement

tfarago:
E.g. `scipy/spatial/_geometric_slerp.py` failed to import.

TODO:

- [x] document that session names cannot use dashes anymore (this is perfectly valid and consistent with Python, sessions are supposed to be importable, so I don't know why we ever allowed such invalid naming in the first place)
- [x] change sessions like `online-reconstruction` to `online_reconstruction` in examples

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -5,7 +5,7 @@ omit =
     concert/devices/cameras/pco.py
     concert/ext/*.py
     concert/processes/ufo.py
-    concert/tests/util/_*.py
+    concert/tests/util/_*
     */python?.?/*
     */lib-python/?.?/*.py
     */site-packages/*
--- a/bin/concert
+++ b/bin/concert
@@ -16,7 +16,7 @@ import zipfile
 import concert
 import concert.config
 import concert.session.management as cs
-from concert._aimport import AsyncMetaPathFinder, eval_nodes
+from concert._aimport import AsyncMetaPathFinder, eval_source
 from concert.ext.cmd import plugins
 
 
@@ -405,9 +405,7 @@ class StartCommand(SubCommand):
 
         if non_interactive:
             with open(path, "rb") as f:
-                source = f.read()
-            nodes = ast.parse(source).body
-            eval_nodes(nodes, {}, filename=path)
+                eval_source(f.read(), {}, filename=path)
         else:
             self.run_shell(path=path, session=session)
 
--- a/bin/concert
+++ b/bin/concert
@@ -1,6 +1,5 @@
 #! python
 
-import ast
 import asyncio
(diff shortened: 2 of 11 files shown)


tfarago:
If these things keep popping up, make the async importer only work on user modules. Perhaps it's not such a bad idea for already now...

tfarago:
@MarcusZuber could you give this a quick spin? I use the async importer more conservatively which hopefully won't break anything.

MarcusZuber:
I tested it for 5 min, since importing libtiff and tiffile did not work without it. So far everithing worked out.

tfarago:
OK, I will re-check it in 2 weeks and merge.

MarcusZuber:
I used this a while "in production" and did not encounter a new issue.

MarcusZuber (approved):

MarcusZuber (changes requested):
I found another feature, that I would like to have (sorry, that I realized this so late).

I use modules for globally organizations of device definitions and instance creations (including top level await, if possible). 
Would it be possible to check a flag of each module to also load it with the async loader (e.g. MODULE.__top_await == True)?

MarcusZuber:
You mean to put this in the session?
```python
import my_package_with_toplevel_await
_ASYNC_IMPORT_PATH.extend(my_package_with_toplevel_await.__path__)
```

tfarago:
Modules in packages should be handled correctly as long as the package is in concert directory. And after this change, if still needed, in the directory which you register.

MarcusZuber:
I have them installed in my global site-package location...
So a module looks like this
mylab/device_instances/motors.py (+ __inti__.py, setup.py etc)

```python
from somewhere import LinearMotor

sample_x = await LinearMotor(...)
```

To get this to work I have to do, or?
```python
from concert._aimport import register
import mylab
register(mylab.__path__)

from mylab.device_instances import sample_x
```

MarcusZuber:
Could I add the register in my mylab/__init__.py, or would this make a knot in the importer?

tfarago:
Once you load a module it's cached, maybe the `import from` would not be affected but it's an obfuscated workaround. Alternatively, at the beginning of your program you can look into `sys.path` and register the necessary ones.

MarcusZuber:
Ok. Let's think about this in the future an proceed with the pr. I will put this on my private to do list as "test if async import works with Marcus crazy python setup".

MarcusZuber (approved):

tfarago:
Sounds good. I wanted to rip `aimport` off of concert to a standalone repository, but first we need the decentralization, so keep this in mind and when it's on its own I think it will get easier.

PR #500, opened 2022-07-01
Do not allow async import for Python < 3.8

tfarago:
We can either do this or bump the minimum python version. If the former:

- [ ] disable non-interactive top level await for Python < 3.8


MarcusZuber:
Do we still need this? Even 3.8 is outdated.

tfarago:
I think not, but I will bump the minimum Python version to 3.8 then.

Issue #302, opened 2014-08-25
Writing meta data


matze:
At this time the only "meta data" that we write are logs of the experiments. This may not suffice for two reasons:
1. The data is unstructured and most likely not very machine readable because that largely depends on who calls `log.info`.
2. The current state of the devices is not recorded if it wasn't set by the user before. Although it would be good practice to do that every time an experiment is started, we can assume that this won't be the case all the time.

However, any mechanism that will be resolve this issue largely depends on what data format we will finally decide on (flat files vs. HDF5).

PR #501, opened 2022-07-08
addon for checking pco timestamps

MarcusZuber:
Addon, that checks the frame-number from the timestamp for pco-cameras.


MarcusZuber:
I only have to check with the real hardware if the numbering in the Timestamp starts with zero or one.

concert/experiments/addons.py
```diff
@@ -362,6 +363,36 @@ async def _write_single_image(self, name, image):
...
+    def _detach(self):
+        """Detach all acquisitions."""
+        for acq in self.acquisitions:
+            acq.consumers.remove(self._timestamp_checks)
+
+    async def _check_timestamp(self, producer):
+        self.timestamp_incorrect = False
+        i = 0
+        async for item in producer:
+            timestamp = Timestamp(item)
```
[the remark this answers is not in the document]
MarcusZuber: I'm already experimenting with this. If the `camera.convert` only applies numpy-array-views I can handle it.
Do you think we have to make this robust enough against other stuff?

concert/experiments/addons.py
```diff
@@ -362,6 +363,31 @@ async def _write_single_image(self, name, image):
...
+
+    def _detach(self):
+        """Detach all acquisitions."""
+        for acq in self.acquisitions:
+            acq.consumers.remove(self._timestamp_checks)
+
+    async def _check_timestamp(self, producer):
+        async for i, item in enumerate(producer):
+            image = item.base
+            timestamp = Timestamp(image)
```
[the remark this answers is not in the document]
MarcusZuber: Should we log this for every failed image or stop the addon after the first frame (in an Acquisition) without a valid timestamp? (See: TODO)

concert/experiments/addons.py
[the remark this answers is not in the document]
MarcusZuber: Do we expect more crazy convert functions, than what I test now?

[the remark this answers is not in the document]
tfarago: No, I think we need a reasonable compromise, i.e. be able to deal with the things we use, so numpy stuff; and if we cannot just write it once to log and stop trying the next images in the stream.
Alternatively, we could forbid `convert` to be any function, but we could make a `Transformation` class with two methods: `forward` and `backward`. Then we could let people implement new transformations.
Which do you prefer?

concert/experiments/addons.py
```diff
@@ -385,7 +385,11 @@ async def _check_timestamp(self, producer):
         self.timestamp_incorrect = False
         i = 0
         async for item in producer:
-            timestamp = Timestamp(item)
+            if item.base is None:
+                img = item
+            else:
+                img = item.base
+            timestamp = Timestamp(img)
```
tfarago: Clever, I like it. You could just save a lot of space by `timestamp = Timestamp(item if item.base is None else item.base)`. To me that's also easily readable, but it's also personal taste I guess.

concert/experiments/addons.py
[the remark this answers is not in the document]
MarcusZuber: I have an idea that is in-between both. I will push it when I have some time.

MarcusZuber:
What I did:

I create a subclass of numpy.ndarray, that can hold the original frame.
If the convert-function if a view, then it holds only a reference to itself.
If the convert function creates a new array it holds a reference to the original frame.

I don't expect any major performance issues from this, since these are only references. The only difference is, that the orginal_image will get out of scope later and therefore it can stany longer in memory. As long as there aore only views used, there should not be any difference.

concert/tests/integration/test_pcotimestamp_addon.py
```diff
@@ -0,0 +1,160 @@
...
+                timestamp = Timestamp(img)
+                self.assertEqual(timestamp.number, i + 1)
+
+    async def run_test_addon(self):
+        """
+        Tests PCOTimestampCheck addon.
+        """
+        await self.camera.set_random_timestamp_numbers(False)
+        await self.exp.run()
+        self.assertFalse(self.addon.timestamp_incorrect)
```
tfarago, line 127: OK, and does the Acquisition finish? I am not completely sure what happens if you `break` in a consumer.

concert/experiments/addons.py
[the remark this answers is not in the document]
tfarago: I think I answered this elsewhere and it was implemented.

concert/tests/integration/test_pcotimestamp_addon.py
MarcusZuber: I think so, but I will check it once more 😃

MarcusZuber:
I did some performance-tests with dummy devices (`concert.devices.cameras.dummy.Camera` and the one I implemented in the tests here) with the current master-branch and this branch and used them with the `concert.experiments.synchrotron.Radiography` with 3000 projections and 200 flats and darks.
On my local computer there was no significant drop in performance (all runs 67 +- 1s execution time).

Also in the memory usage I didn't see a significant difference when writing to local disc. Since, the write-queue strips the original image, there should also not a major issue when the write-queue is used a lot.

@tfarago : Do you have any other suggestions to test?

MarcusZuber:
If you use a large `Accumulate` coro it could double the amount of required memory if you don't use a numpy view in the convert.

I will add on Monday, that the addon saves the timestamps to disk as well, so you can crop and still store the times. But this will not change something that is already here.

tfarago:
So, `np.transpose` will create a view, all fine. *BUT*, `np.rot90` will create a copy, so people will double the memory without even knowing it, correct? Why do we have to mess with the `cameras.base`, since this is special to pco? I would rather like keeping it simple and not change anything in base, and especially I wouldn't rely on people creating views...

tfarago:
I mean it's very clever with the view and all but I have doubts from the performance point of view and the requirement to know about numpy views. And also, can I actually rotate by 90 deg by a view?

MarcusZuber:
numpy.rot90 creates a view in the cases I tested. But numpy decides by some magic rules if it can do a view or a new instance.

tfarago:
```python
import numpy as np
a = np.arange(4).reshape(2, 2)
ar = np.rot90(a)
at = a.transpose()
a.__array_interface__['data'][0] == at.__array_interface__['data'][0]
a.__array_interface__['data'][0] == ar.__array_interface__['data'][0]
```
Already a pretty simple attemt fails to make just a view...

MarcusZuber:
> So, `np.transpose` will create a view, all fine. *BUT*, `np.rot90` will create a copy, so people will double the memory without even knowing it, correct? Why do we have to mess with the `cameras.base`, since this is special to pco? I would rather like keeping it simple and not change anything in base, and especially I wouldn't rely on people creating views...

The stuff in the camera and coroutine base classes is general enough, that it can be used with any camera. But we can discuss later what we want to have.

MarcusZuber:
> ```python
> import numpy as np
> a = np.arange(4).reshape(2, 2)
> ar = np.rot90(a)
> at = a.transpose()
> a.__array_interface__['data'][0] == at.__array_interface__['data'][0]
> > True
> a.__array_interface__['data'][0] == ar.__array_interface__['data'][0]
> > False
> ```
> 
> Already a pretty simple attemt fails to make just a view...

But the `ar.base` and `at.base` hold a reference to the instance of `np.arange(4)` in both cases. So you have the same memory/gc behaviour like with my implementation.

concert/coroutines/base.py
```diff
@@ -168,6 +171,7 @@ async def feed_queue(producer, func, *args):
         future = loop.run_in_executor(None, func, pqueue, *args)
         prio = 1
         async for item in producer:
+            item = item.view(np.ndarray)
```
MarcusZuber, line 174: This is actually not needed anymore, since the time and number should not make such a big influence on the memory consumption.

concert/devices/cameras/base.py
```diff
@@ -144,8 +144,10 @@ async def trigger(self):
 
     @background
     async def grab(self):
-        """Return a NumPy array with data of the current frame."""
-        return self.convert(await self._grab_real())
+        """Return a concert.helpers.Image array (subclass of numpy-array) with data of the current
```
MarcusZuber, line 147: Needs to be reverted

concert/devices/cameras/pco.py
```diff
@@ -1,112 +1,46 @@
...
+class Camera(UcaCamera):
+    async def __ainit__(self, name="pco", params=None):
+        await super().__ainit__(name=name, params=params)
 
-    async def stream(self):
-        """
-        stream()
+    async def grab(self, index=None):
+        image = await self._grab_real(index=index)
+        if await self.get_timestamp_mode() in ["both", "binary"]:
```
tfarago, line 18: I think this should be cached? Especially with `net`, this means we every single time ask the server to ask the camera to get the timestamp mode, let's cache this.

concert/experiments/addons.py
```diff
@@ -362,6 +364,53 @@ async def _write_single_image(self, name, image):
...
+            frame_numbers.append(img.frame_number)
+            timestamps.append(img.timestamp)
+            if img.frame_number != i + 1:
+                self._experiment.log.error(
+                    f"Frame {i + 1} had wrong frame number {img.frame_number}.")
+                self.timestamp_incorrect = True
+            i += 1
+        self.write_timestamps_to_disc(frame_numbers, timestamps, directory=current_acquisition_directory)
+
+    def write_timestamps_to_disc(self, frame_numbers: list, timestamps: list, directory: str):
```
tfarago, line 402: It's `disk`

tfarago:
> The commits need to be cleaned up, but I think it now does its job in a much cleaner way.
> 
> In each acquistion-subfolder now a timestamp.json and frame_numbers.json is stored with such a content:
> 
> timestamp: { "0": "2022-09-19 16:32:52.987985", "1": "2022-09-19 16:32:53.010777", "2": "2022-09-19 16:32:53.033412", "3": "2022-09-19 16:32:53.055898", "4": "2022-09-19 16:32:53.078596", "5": "2022-09-19 16:32:53.101142", "6": "2022-09-19 16:32:53.123399", "7": "2022-09-19 16:32:53.146141", "8": "2022-09-19 16:32:53.168262", "9": "2022-09-19 16:32:53.190795" }
> 
> frame_number: { "0": 1, "1": 2, "2": 3, "3": 4, "4": 5, "5": 6, "6": 7, "7": 8, "8": 9, "9": 10 }

That's one way to do it. But I think a cleaner way would be to store the metadata together with the data. That's also how the community is doing it (hdf, nexus...). I would stick with tif for now and just made the writer aware of metadata and just dumped it together with the images. What do you think? Then your Addon will be checking them, what it was originally intended to do and we will have a way of storing *any* kind of metadata, so it's more extensible in the future.

tfarago:
I checked and ImageJ can open it, even though only the first image's metadata is shown but we can figure this out later. With `tifffile` you can retrieve the metadata for every image by e,g. `tifffile.TiffReader('meta-test.tif').pages[123].description`

MarcusZuber:
Sounds all good. I will implement and test it tomorrow.

How does our tiff writer store the images? Are the pages linked lists? If yes, did you maybe check the performance of reading timestamps from a multi tiff? I'm curious how long it would take to read all timestps from a typical acquisition, since I have to do a lot of reference jumps to access all the data.

tfarago:
For our typical data set I get this data on my local hard drive (after clearing caches):
```python
In [23]: %timeit -r 1 -n 1 [reader.pages[i].description for i in range(3000)]
626 ms ± 0 ns per loop (mean ± std. dev. of 1 run, 1 loop each)
```
I think tifffile does it in a clever way and puts all metadata at the end, also the linked list. But if you want to be sure look inside... And practically we can test it on LSDF because there you'd notice the difference between having to go through 32 GB of data or only a tiny portion at the end of the physical file.

MarcusZuber:
With the `tifffile `package I think I got it working. Did you maybe find in `libtiff `how to store metadata?

tfarago:
> With the `tifffile `package I think I got it working. Did you maybe find in `libtiff `how to store metadata?

No idea, but we don't use that anymore and it's easy to just not use them if this writer doesn't support it...

MarcusZuber:
I think with this PR we can finally close https://github.com/ufo-kit/concert/issues/302.

concert/tests/integration/test_pcotimestamp_addon.py
MarcusZuber: Yes is finishes. (Otherwise the` await self.exp.run()` would also not finish).

concert/devices/cameras/pco.py
MarcusZuber: Done in the `start_readout` and `_start_real`.

concert/devices/cameras/pco.py
```diff
@@ -1,112 +1,47 @@
...
+        if self._timestamp_enabled:
+            try:
+                timestamp = Timestamp(img)
+                img = self.convert(img)
+                img = img.view(ImageWithMetadata)
+                img.metadata['frame_number'] = timestamp.number
+                img.metadata['timestamp'] = timestamp.time.isoformat()
+                return img
+            except TimestampError:
+                LOG.error("Can not extract timestamp from frame.")
```
tfarago, line 43: You surely need to re-raise. Otherwise people will be very surprised what's happening. And also tests should cover the re-raise.

concert/experiments/addons.py
```diff
@@ -362,6 +364,44 @@ async def _write_single_image(self, name, image):
...
+                    self._experiment.log.error("No 'frame_number' present in image."
+                                               "camera.timestamp needs to be set to 'both' or"
+                                               "'binary' to use this addon."
+                                               "Works only with pco cameras.")
+                    self.timestamp_incorrect = True
+                    return
+            if img.metadata['frame_number'] != i + 1:
+                self._experiment.log.error(
+                    f"Frame {i + 1} had wrong frame number {img.metadata['frame_number']}.")
+                self.timestamp_incorrect = True
```
tfarago, line 401: So `timestamp_incorrect` could actually mean that there is no timestamp or the order of images is wrong, don't we need to distinguish those? Also, shouldn't an error be raised, at least when the order is wrong? Or it should at least be an option like in C compiler when you turn warnings to errors. I am not sure but for some experiments the order might be crucial I guess...

concert/tests/integration/test_pcotimestamp_addon.py
```diff
@@ -0,0 +1,169 @@
...
+                                     flat_motor=flat_motor,
+                                     shutter=shutter,
+                                     radio_position=0 * q.mm,
+                                     flat_position=10*q.mm,
+                                     num_projections=10,
+                                     num_flats=5,
+                                     num_darks=5)
+        self.addon = PCOTimestampCheck(self.exp)
+        self.writer = ImageWriter(self.exp.acquisitions, self.walker)
+
```
tfarago, line 101: Somewhere here the `TimestampError` re-raising should be checked.

concert/experiments/addons.py
MarcusZuber: I will add the option to raise an Exception at the end of the last acquistion, if the order was not right. I would also raise it if the timestamp is incorrect, since I guess if you attach this addon, you really want to know if the timestamps are there and in the correct order.

concert/devices/cameras/pco.py
MarcusZuber: But re-raising it would mean, that every frame with a broken timestamp would be lost. Do we want this?

concert/tests/integration/test_pcotimestamp_addon.py
```diff
@@ -0,0 +1,169 @@
...
+        time = datetime.now()
+        if not await self.get_random_timestamp_numbers():
+            img = add_binary_timestamp(img, time, self.number)
+        else:
+            img = add_binary_timestamp(img, time, np.random.randint(1, 1000))
+        self.number += 1
+        return img
+
+
+PCOCamera._grab_real = Camera._grab_real
```
tfarago, line 80: Or you could derive from the `DummyCamera` first _and_ `PCOCamera` second and override `_grab_real`. That way things like `start_recording` would be done on the dummy (left-right MRO). Basically, you can choose what parent gets called when, which is exactly what you need.

tfarago (commented):
Almost perfect.

concert/experiments/addons.py
MarcusZuber: See https://github.com/ufo-kit/concert/pull/501/commits/77644755a46c72c50e2c32fe4375fce0b0789b41. The tests need to adjusted accordingly.

MarcusZuber:
When rewriting the test I found this:

https://github.com/ufo-kit/concert/blob/75c881f6810fab4155180b18df3093a447f8903e/concert/experiments/base.py#L298

Is there a reason why always the `StateError` and not just `e` is raised? I can then not catch the error simply by its type, but have to check the message.

tfarago:
I think historically we did this because `StateError` used to set the state to error. That is a weird solution when I think about it now but on the other hand, it will be gone with #481, take a [look](https://github.com/ufo-kit/concert/blob/decentralize/concert/experiments/base.py#L428-L431). The only thing left from the original exception handling is the finally block. There we can also change the `StateErorr` by the way, but it's not related to your issue here. So, maybe it makes sense you rebase on the #481 because I don't think much more will change there. I just need to write a bunch of tests basically.

concert/devices/cameras/pco.py
tfarago: It would mean the addon stops the whole data acquisition, which is what you might want if time stamp is super-important.

tfarago:
I just wanted to add `contiguous=True` flag to the tif writer because it writes the IFDs at the end of the file, which may be important for slow storage. Someone told this to me once, although I have never had problems... Bad thing is that with this flag, our metadata will be written only for the first image, which breaks this PR... So we either choose performance of reading IFDs and need to re-think how to store metadata or leave performance for the future.

concert/devices/cameras/pco.py
tfarago: I explicitly set the timestamp to be there, but then it's not and I don't get an exception. You sure that's correct behavior?

tfarago: Also the whole try block is too long, the exception is only about the timestamp, so why is everything there? A lot can move above I think, you just add or not add the metadata based on the `_timestamp_enabled` flag.

MarcusZuber: OK, now the TimestampError will be re-raised in the grab, and everithing else is moved out from the try-block.

concert/readers.py
```diff
@@ -114,7 +123,18 @@ def _get_num_images_in_file_real(self):
         return len(self._file.pages)
 
     def _read_real(self, index):
-        return self._file.pages[index].asarray()
+        image = self._file.pages[index].asarray().view(ImageWithMetadata)
+        if self._metadata_file:
+            image.metadata = self._json_metadata[str(index)]
+        else:
+            try:
+                image.metadata = ast.literal_eval(self._file.pages[index].description)
```
tfarago, line 131: `self._file.pages[index].description` doesn't return a dict? I am wondering why you need the `literal_eval`.

MarcusZuber: No. It returns a string, that is formated like a dictionary assignment. This fits so well with my other tiff experiences...

MarcusZuber: Or do you think I do something wrong when writing the metadata? The ImageDescription from the tests are shown like this in ImageJ "ImageDescription: {"time": "2022-10-14T12:25:37.739517", "random_value": 0.9344031366187244, "shape": [10, 10]}". And the shape is automatically added by tifffile. So I guess the writing is correct like this.

tfarago (approved):
As an intial implementation it's fine, we'll deal with performance later.

tfarago:
Please fix flak8 before hitting the button.

PR #502, opened 2022-07-22
Fix get state from awaitable

tfarago:


concert/helpers.py
```diff
@@ -299,7 +299,13 @@ async def get_state_from_awaitable(awaitable) -> str:
...
         return 'running'
-    elif task.exception() is None:
-        return 'standby'
     else:
-        return 'error'
+        try:
+            exc = task.exception()
+        except asyncio.CancelledError:
+            # That's the only exception type left which can be raised
+            return 'error'
```
tfarago, line 307: Here it's up for discussion, I can find arguments for making this 'standby' or a completely new state 'cancelled'. If it won't be standby we need to update the `check` targets of `Experiment` and `Acquisition` `run` methods.

MarcusZuber: I would prefere 'cancelled'. 
I can directly add the new state in https://github.com/ufo-kit/concert/pull/504 to the checks.

MarcusZuber (changes requested):
If you agree to add the "cancelled" state and add it, I'm fine with it.

MarcusZuber:
Down really need to set the self._state_value?
We always use the helper function to get the state.

tfarago:
> Do we really need to set the self._state_value? We always use the helper function to get the state.

Done.

concert/experiments/base.py
```diff
@@ -70,7 +70,7 @@ async def _run(self):
         await asyncio.gather(*coros, return_exceptions=False)
 
     @background
-    @check(source=['standby', 'error'], target='standby')
+    @check(source=['standby', 'error'], target=['standby', 'cancelled'])
```
MarcusZuber, line 73: "cancelled" should also be in source, otherwise we can  not re-run an experiment or acquisition, if it was cancelled before.
In https://github.com/ufo-kit/concert/pull/504/commits/6c55a59021660c269966b19ca43af1e2c1b6426d I added them (also to all the other places in the experiments).
About the target I'm not 100% sure if we add it there or raise the Exception, that the final state is not correct.

MarcusZuber (changes requested):
I think we need the cancelled in the source-state-list. Could you maybe add a test, that cancels and then tries to re-runs the experiment?

concert/experiments/base.py
tfarago: Right, actually it should raise the exception. I just swallow it now but when an experiment is used in a director it must notify it. I falsely assumed the experiment is at the top of the food chain, so I'd change that actually to raise the `CancelledError`, ok?

tfarago:
> I think we need the cancelled in the source-state-list. Could you maybe add a test, that cancels and then tries to re-runs the experiment?

That's fixed with the re-worked 971bf4ed111d3e03e3f1c5edcc547810139ccee3. I also check the cancelled state in a nicer way.

tfarago:
Alright, please give it one more look and hit the button if you like it.

tfarago:
:fireworks:

PR #503, opened 2022-08-04
Sets the ipython-session profile to "concert_*session_name*"

MarcusZuber:
This creates and uses independent ipython session profiles for each concert session.
This helps to separates histories between sessions.

TODOs:

- [x]  Delete profile folders together with the sessions
- [x] Test profile creation
- [x] Test profile deletion
- [x] Change concert.log to a rotating format

### DIFF
--- a/bin/concert
+++ b/bin/concert
@@ -406,6 +406,8 @@ class StartCommand(SubCommand):
         print("Welcome to Concert {0}".format(concert.__version__))
 
         ip_config = traitlets.config.Config()
+        if session is not None:
+            ip_config.BaseIPythonApplication.profile = f"concert_{session}"
         if path and path.endswith('.py'):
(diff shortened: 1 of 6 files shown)


tfarago:
What does this actually do? Creates a profile for every session or just dumps history somewhere else?

MarcusZuber:
It creates one profile per Session (name).

tfarago:
I wonder if the session history should be cleaned up at some point. I can imagine `concert rm session` would delete the profile, what do you think?

MarcusZuber:
To delete them makes sense. I will add this.
I will also have a look if there is a way to somehow remove old histories (maybe like log rotate or similar...).

MarcusZuber:
After https://github.com/ufo-kit/concert/pull/552 is merge I can add the tests (I need access to the Commands)

PR #504, opened 2022-08-08
Some updates for the imaging experiments

MarcusZuber:
In the first commit I create a bit more code duplications, but this allows to do the actions in the _prepare_radios and _prepare_flats in a better order.

### DIFF
--- a/concert/experiments/imaging.py
+++ b/concert/experiments/imaging.py
@@ -10,6 +10,8 @@ from concert.base import background, Parameter, Quantity, Parameterizable, \
     AccessorNotImplementedError
 from concert.base import check
 
+from concert.experiments.base import _runnable_state
+
 
 async def frames(num_frames, camera, callback=None):
     """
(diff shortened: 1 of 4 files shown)


MarcusZuber:
The state-checks we have to synchronize with https://github.com/ufo-kit/concert/pull/502, but the rest is open for discussion.
I did this mostly to make it easier to create sub-classes without rewriting the full `take_radios`.

tfarago:
> The state-checks we have to synchronize with #502, but the rest is open for discussion. I did this mostly to make it easier to create sub-classes without rewriting the full `take_radios`.

Let's keep this about cleaning the actual DAQ things and #502 will take care of the state.

MarcusZuber:
> > The state-checks we have to synchronize with #502, but the rest is open for discussion. I did this mostly to make it easier to create sub-classes without rewriting the full `take_radios`.
> 
> Let's keep this about cleaning the actual DAQ things and #502 will take care of the state.

The check fix is now only present for the imaging-experiments. For the base experiment and acquisition we use the commit from #502.

concert/experiments/imaging.py
```diff
@@ -736,30 +750,8 @@ async def _finish_radios(self):
...
-            await self._prepare_radios()
-            await self._camera.set_trigger_source("SOFTWARE")
-            async with self._camera.recording():
-                for i in range(num_projections):
-                    await self._prepare_frame(i)
-                    await self._camera.trigger()
-                    yield await self._camera.grab()
-        finally:
-            await self._finish_radios()
-
```
tfarago, line 744: both `_take_radios` slipped into this commit I guess.

tfarago (commented):
A bad squash I guess. Otherwise this looks good.

concert/experiments/imaging.py
MarcusZuber: Do you like this commit msg more?

PR #506, opened 2022-11-17
Minor tango-connection changes

MarcusZuber:
The import should be tango instead of PyTango since a while (but PyTango still works)

The tango_host can also be specified in the uri of the DeviceProxy.


MarcusZuber:
I will test this later today with real hardware.

tfarago (approved):

PR #510, opened 2023-01-23
Install parameters on instance instead of class

MarcusZuber:
This should attach the parameters to the single instances and not to the class.

First tests are working:

```python
from concert.base import Quantity

l = await LinearMotor()
l2 = await LinearMotor()
l2.install_parameters({"pos2": Quantity(q.deg)})

l
<concert.devices.motors.dummy.LinearMotor object at 0x7f10a94d39a0>
 Parameter        Value                        
 motion_velocity  200000.0 millimeter / second 
 position         5 millimeter                 
 state            standby  

l2
<concert.devices.motors.dummy.LinearMotor object at 0x7f10a89d6fb0>
 Parameter        Value                        
 motion_velocity  200000.0 millimeter / second 
 pos2             N/A                          
 position         0 millimeter                 
 state            standby  
```

The actual use-case of this is, when employing different uca-cameras.
With the class-wide implementation, each of the `concert.devices.camera.uca.Camera` adds all plugin-parameters to the uca.Camera class.

### DIFF
--- a/concert/base.py
+++ b/concert/base.py
@@ -336,21 +336,21 @@ class Parameter(object):
     @memoize
     def setter_name(self):
         if self.fset:
-            return self.fset.__name__
+            return self.fset.__class__.__name__
 
         return '_set_' + self.name
 
     @memoize
     def getter_name(self):
         if self.fget:
-            return self.fget.__name__
+            return self.fget.__class__.__name__
 
         return '_get_' + self.name
 
     @memoize
     def getter_name_target(self):
         if self.fget_target:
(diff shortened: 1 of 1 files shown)


MarcusZuber:
@tfarago : Do you see any general problems here? I will also add a few tests for this later.

tfarago:
What is the difference between `setattr(self, 'foo', foo)` and `self.__dict__['foo'] = foo`? You changed it to this kind and I am curious what the benefits are.

PR #516, opened 2023-01-31
Mapping experiment

MarcusZuber:
Experiments for mapping of a sample.


tfarago:
I see how this is useful, but a similar remark to the one about exper. simplification: wouldn't it make sense to rather implement it via our own scanning routines? Theyare quite flexible and the treajectories could be just passed to a scan.

MarcusZuber:
I thought about using the scans, but for what I'm planing in particular (after this more or less abstract implementation here) more interaction between the experiment and the scan is required.

tfarago:
OK, do your thing, then I'll review, but again keep in mind #481 please, which in this context means: do not write more code then necessary for the near future please.

PR #518, opened 2023-07-18
Fix imagej

tfarago:


tfarago:
Why does libtiff installation fail now? Oh man, the setting up of actions breaks more often than the code itself.

MarcusZuber (approved):
The issue is fixed.

The only thing I don't like so much (but is only slightly related to the process-spawning issue): We write big-tiff, which requires the bioformats-plugin and you then also have to click trough the bioformats-import dialogues for just viewing a single image.

MarcusZuber:
It also always a new imagej instance. 
When imagej starts it tell you the "imagej instance id", which can be used for opening future images in the same imagej instance. 
```python
concert > imagej(img)

concert > Open other images in this ImageJ panel as follows:
  imagej -p 1 <image1> [<image2> ... <imageN>]

```

tfarago:
On my machine this also doesn't happen. Update imagej?

PR #522, opened 2023-07-31
chore: Enhancement proposals on decentralize branch

sarkarchandan:
**Premise**: With this PR we would like to make some enhancement proposals on top of the changes in the decentralize branch. The details of the proposed enhancement would be entered gradually and finalized post discussion.


sarkarchandan:
It does not make sense to refactor the api-layer without being familiar with the big picture of the implementation.

PR #523, opened 2023-08-10
chore: Make decentralize design documents and test cases

sarkarchandan:
**Premise**: With this PR we propose to start integrating static type-checking and type-annotations to concert in incremental steps. We also plan to add a design document for the framework which should evolve with the implementation.


sarkarchandan:
The failure in Python 3.8 build is due to the missing coverage file, which is fixed with the PR #512 in master. Once we refresh the decentralize with master or cherry-pick the changes it would be addressed.

PR #525, opened 2023-10-06
Fix timestamp checks for PCO cameras

MarcusZuber:
Fixes timestamp checks for PCO cameras.

This also ignores the index of the first frame and checks only if all subsequent frames are correctly aligned

### DIFF
--- a/concert/devices/cameras/pco.py
+++ b/concert/devices/cameras/pco.py
@@ -15,16 +15,16 @@ LOG = logging.getLogger(__name__)
 class Camera(UcaCamera):
     async def __ainit__(self, name="pco", params=None):
         self._timestamp_enabled = False
-        await super().__ainit__(name=name, params=params)
+        await UcaCamera.__ainit__(self, name=name, params=params)
 
     async def _record_real(self):
-        self._timestamp_enabled = await self.get_timestamp() in ['both', 'binary']
-        await super()._record_real()
+        self._timestamp_enabled = (await self.get_timestamp_mode()).value_name in ['UCA_PCO_CAMERA_TIMESTAMP_BINARY', 'UCA_PCO_CAMERA_TIMESTAMP_BOTH']
+        await UcaCamera._record_real(self)
 
     @background
     async def start_readout(self):
-        self._timestamp_enabled = await self.get_timestamp() in ['both', 'binary']
-        await super().start_readout()
+        self._timestamp_enabled = (await self.get_timestamp_mode()).value_name in ['UCA_PCO_CAMERA_TIMESTAMP_BINARY', 'UCA_PCO_CAMERA_TIMESTAMP_BOTH']
+        await UcaCamera.start_readout(self)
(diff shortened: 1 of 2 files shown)


concert/experiments/addons.py
```diff
@@ -409,6 +412,7 @@ async def _check_timestamp(self, producer):
             raise PCOTimestampCheckError("Not all images contained timestamps.")
 
 
+
```
tfarago, line 415: oops.

concert/devices/cameras/pco.py
```diff
@@ -15,16 +15,16 @@
...
-        await super()._record_real()
+        self._timestamp_enabled = (await self.get_timestamp_mode()).value_name in ['UCA_PCO_CAMERA_TIMESTAMP_BINARY', 'UCA_PCO_CAMERA_TIMESTAMP_BOTH']
+        await UcaCamera._record_real(self)
 
     @background
     async def start_readout(self):
-        self._timestamp_enabled = await self.get_timestamp() in ['both', 'binary']
-        await super().start_readout()
+        self._timestamp_enabled = (await self.get_timestamp_mode()).value_name in ['UCA_PCO_CAMERA_TIMESTAMP_BINARY', 'UCA_PCO_CAMERA_TIMESTAMP_BOTH']
+        await UcaCamera.start_readout(self)
```
tfarago, line 27: Why move away from `super()`?

concert/experiments/addons.py
```diff
@@ -398,9 +399,11 @@ async def _check_timestamp(self, producer):
                                                "Works only with pco cameras.")
                     self.timestamp_missing = True
                     return
-            if img.metadata['frame_number'] != i + 1:
+            if i == 0:
```
tfarago, line 402: You have `if i == 0` above, why repeat the same if?

concert/experiments/addons.py
```diff
@@ -398,9 +399,11 @@ async def _check_timestamp(self, producer):
...
+                first_frame = img.metadata['frame_number']
+            if img.metadata['frame_number'] != i + first_frame:
```
tfarago, line 404: Is it safe to rely on the first frame index being 1? All PCO's behave the same way for sure?

concert/experiments/addons.py
```diff
@@ -388,6 +388,7 @@ async def _check_timestamp(self, producer):
         self.timestamp_missing = False
         i = 0
         last_acquisition = await self._experiment.acquisitions[-1].get_state() == "running"
```
tfarago, line 390: I know this isn't new but it's really a hack, we should improve this in the future.

concert/devices/cameras/pco.py
tfarago: Still, why did you change this?

tfarago (commented):
Just out of curiosity, why did you need to change the `super()` call? It didn't work or what?

PR #528, opened 2023-12-07
storage: enable directory rights specification

tfarago:
one can now do:
```python
from concert.storage import DirectoryWalker
walker = DirectoryWalker(rights=0o0755)
walker.descend("foo")
```
to see this:
```bash
$ ls -l
drwxr-xr-x 2 tomas tomas 4096 Dez  7 10:53 foo
```

### DIFF
--- a/concert/storage.py
+++ b/concert/storage.py
@@ -70,18 +70,20 @@ def write_libtiff(file_name, data):
     return file_name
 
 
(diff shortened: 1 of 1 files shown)


MarcusZuber (approved):
If you note somewhere that we also add this to the isngle files to make it consistent I'm fine with it

PR #529, opened 2023-12-15
Extend experiment logging

MarcusZuber:
Unfortunately, there are often devices that will not work always... Therefore, I added the optional flag to the devices to log, so I can add a device to log, that is nice-to-have, but will not stop the experiment execution if it malfunctions.

I added also a second log at the end of the experiment execution. I think i is useful, but this is open for discussions.

I added a prepare and a finish function. Some devices need to be in a specific state to get useful data. In this function the user can handle such things.

### DIFF
--- a/bin/concert
+++ b/bin/concert
@@ -12,6 +12,8 @@ import signal
 import subprocess
 import tempfile
 import zipfile
+from logging.handlers import TimedRotatingFileHandler
+
 import concert
 import concert.config
 import concert.session.management as cs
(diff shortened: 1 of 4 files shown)


MarcusZuber:
[cf933cb](https://github.com/ufo-kit/concert/pull/529/commits/cf933cb84dce8260f21b0c89232a4b5ec293bd07) Is a simple implementation of a json logger that allows to log stuff with the time as a key.

PR #531, opened 2024-01-18
concert: Fix merge errors

sarkarchandan:
**Premise**: Fixes the errors introduced during merge of decentralized to master. Verified with session tests.

### DIFF
--- a/concert/storage.py
+++ b/concert/storage.py
@@ -91,27 +91,37 @@ def write_libtiff(file_name, data):
     return file_name
 
 
-def create_directory(directory, rights="750"):
-    """Create *directory* and all paths along the way if necessary. *rights* are a string
-    representing a combination for user, group, others.
+def create_directory(directory: str, rights: int = 0o0750) -> None:
+    """
+    Create *directory* and all paths along the way if necessary
+
+    :param directory: directory to be created
+    :type directory: str
+    :param rights: access rights for the directory, defaults rwx for
+        user and rx for the group
+    :type rights: int
     """
     if not os.path.exists(directory):
         LOG.debug("Creating directory {}".format(directory))
-        os.makedirs(directory, int(rights, base=8))
+        os.makedirs(name=directory, mode=rights)
 
 
-def write_images(pqueue, writer=TiffWriter, prefix="image_{:>05}.tif", start_index=0,
-                 bytes_per_file=0, rights="750"):
+async def write_images(producer: AsyncIterable[ArrayLike],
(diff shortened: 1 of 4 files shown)


sarkarchandan:
Added `master-with-decentralized` to test workflow to reveal problems, of which there are a few. Some tests are broken during merge. One reason is that previously `addon` was a module inside `concert.experiments` package and types e.g., `PCOTimestampCheck` or `PCOTimestampCheckError` were defined in it. Now `addons` is a sub package in `concert.experiments` and types are missing. As a result some tests are broken.

concert/storage.py
```diff
@@ -91,27 +91,37 @@ def write_libtiff(file_name, data):
     return file_name
 
 
-def create_directory(directory, rights="750"):
```
tfarago, line 94: It should be this, not 0o0750.

concert/storage.py
```diff
@@ -91,27 +91,37 @@ def write_libtiff(file_name, data):
...
+        os.makedirs(name=directory, mode=rights)
 
 
-def write_images(pqueue, writer=TiffWriter, prefix="image_{:>05}.tif", start_index=0,
-                 bytes_per_file=0, rights="750"):
+async def write_images(producer: AsyncIterable[ArrayLike],
+                       writer: Type[TiffWriter] = TiffWriter,
+                       prefix: str = "image_{:>05}.tif",
+                       start_index: int = 0,
+                       bytes_per_file=0) -> int:
```
tfarago, line 113: Rights should make it back into the parameters.

concert/storage.py
```diff
@@ -125,7 +135,7 @@ def write_images(pqueue, writer=TiffWriter, prefix="image_{:>05}.tif", start_ind
         im_writer = writer(prefix, bytes_per_file, append=True)
 
     if dir_name and not os.path.exists(dir_name):
-        create_directory(dir_name, rights=rights)
+        create_directory(dir_name)
```
tfarago, line 138: We want the rights here as well.

concert/storage.py
```diff
@@ -369,7 +379,7 @@ async def __ainit__(self,
 
     async def _descend(self, name: str) -> None:
         new = os.path.join(self._current, name)
-        create_directory(new, rights=self._rights)
+        create_directory(new)
```
tfarago, line 382: here too.

tfarago (changes requested):
Look at #528 regarding the rights. There I changed the int-based specification to string-based one, which you reverted here.

PR #533, opened 2024-03-21
Proper implementation of 2D-slits

MarcusZuber:
TODOs:
- [x] Allow multiple states per Parameterizable
- [ ] Add tests for multiple states
- [ ] Add some more documentation for the states
- [ ] Add slit implementations
   - [X] Slit with 4 motors, each moving one blade
   - [ ] Slit with motors moving the offset and gap
- [ ] Add tests for slits

About the multiple states:
We use multiple devices, that are composed of sub-devices, where the state-machine can be split in multiple state-machines. Therefore, I introduced the possibility to use multiple states.

```python
class ComplicatedDevice(Device):
   state_1 = State(default="standby")
   state_2 = State(default="standby")
   parameter_1 = Quantity(q.mm, check("standby", state_name="state_1")
   parameter_2 = Quantity(q.mm, check("standby", state_name="state_2")

   @transition(immediate="moving", target="standby", state_name="state_1")
   async def _set_parameter_1(self, p):
       # Do stuff

   @transition(immediate="moving", target="standby", state_name="state_2")
   async def _set_parameter_2(self, p):
       # Do stuff
```

### DIFF
--- a/concert/base.py
+++ b/concert/base.py
@@ -210,19 +210,20 @@ class LockError(Exception):
     pass
 
 
-def transition(immediate=None, target=None):
+def transition(immediate=None, target=None, state_name='state'):
     """Change software state of a device to *immediate*. After the function execution finishes
     change the state to *target*. On :py:class:`.asyncio.CancelledError`, state is set to *target*
     and cleanup logic must take place in the callable to be wrapped.
+    *state_name* can be used if multiple states are used.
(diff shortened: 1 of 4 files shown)


MarcusZuber:
Tests fail due to race-condition when motion is started. Have to find a elegant solution for this.

PR #540, opened 2024-07-26
Stuff after decentralized

MarcusZuber:
Here I put everything I need at the beamlines to work with the decentralized and all things that piled up while decentalized was developed.

Will be cleaned up by me when decentralized is finished.

Changes:
- Experiement.add_device_to_log now has a third parameter "optional". A device that is added with optional=True will be ignored when reading its properties throws an exception
- experiment.json is now created at the beginning and end of the experiment (creating experiment_start.json and experiemtn_finish.json)
- with Experiment.log_devices_at_start/finish both json-logs can be switched on and off.
- The `concert.base.AsyncMeta` inherits from `abc.ABCMeta`, so abstract base classes can be properly used for everything.
-  All base classes should use now `@abc.abstractmethod `insted of raising an `NotImplemeted` exception.
- The monochromator that sets/gets the wavelength from the pyhsical device was removed. I never saw a monocromator where you set the wavelength. It is still possible to implement it (but is not in the base implementation)
- Compatibility with numpy 2.0.X
- Allow the use of multiple states. Some devices have multiple independent state machines and actions can be dependent on different subsets of the states of these state machines. E.g. an X-ray tube has a vacuum system, one ore multiple cooling systems and a safety interlock. With this I could do this:
```python
class XRayTube:
   interlock = State()
   cooling = State()
   vacuum = State()
   state = State()

  @check(source="off", target="on")
  @check(source="running", state_name="cooling")
  @check(source="good", state_name="vacuum")
  @check(source="closed", state=name="interlock")
  async def on(self):
    # switch xrays on
```
  instead of using parameters and throwing exceptions and/or ugly concatenated states.

I would drop 91cb627377f0f36a78ff5c8c9dd2eb97aebccb5e here and think of a more elegant way to get this behaviour.


bin/concert
```diff
@@ -23,6 +23,19 @@ from concert.ext.cmd import plugins
...
+except Exception as e:
+    import os
+    import signal
+    import sys
+    import traceback
+    print("\\nAn error occured while starting session `{0}':")
+    print("-------------------------------------------" + "-" * len('{0}'))
+    print(traceback.format_exc(), file=sys.stderr)
+    os.kill(os.getpid(), signal.SIGTERM)
+"""
```
tfarago, line 38: @MarcusZuber do we want to remove the lock file if this happens?

concert/base.py
```diff
@@ -1180,7 +1185,7 @@ async def __ainit__(self):
         pass
 
 
-class Parameterizable(AsyncObject):
+class Parameterizable(AsyncObject, abc.ABC):
```
tfarago, line 1188: @MarcusZuber we are 10000 % sure this and the ABCMeta doesn't break anything, yes?

concert/devices/cameras/base.py
```diff
@@ -108,12 +103,17 @@ class Camera(Device):
     frame_rate = Quantity(1 / q.second, help="Frame frequency")
     trigger_source = Parameter(help="Trigger source")
     _senders: Dict[CommData, ZmqSender]
+    mirror = Parameter(help="Mirror the image")
+    rotate = Parameter(help="Rotate the image")
```
tfarago, line 107: Why not _one_ parameter with choices:
```python
convert = Choices("fliplr", "flipud", "rot90", ...)
```

tfarago: If we keep adding things we will have maaaaany parameters.

concert/devices/cameras/base.py
```diff
@@ -108,12 +103,17 @@ class Camera(Device):
...
 
     async def __ainit__(self):
+        self._rotate = 0
+        self._mirror = False
         await super(Camera, self).__ainit__()
-        self.convert = identity
         self._grab_lock = asyncio.Lock()
         self._senders = {}
+        #await self.set_mirror(False)
+        #await self.set_rotate(0)
```
tfarago, line 116: comments should be removed

concert/devices/cameras/base.py
```diff
@@ -268,23 +273,44 @@ async def register_endpoint(self, endpoint: CommData) -> None:
             sndhwm=endpoint.sndhwm
         )
 
+    async def unregister_all(self) -> None:
+        self._senders = {}
```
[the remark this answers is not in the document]
tfarago: @sarkarchandan you please check this too.

concert/devices/cameras/uca.py
```diff
@@ -267,6 +267,8 @@ async def __ainit__(self, params=None):
         self._ucad_host = self.uca.props.host
         self._ucad_port = self.uca.props.port
         weakref.finalize(self, _ucad_unregister_all, self._ucad_host, self._ucad_port)
+        if ":" in self._ucad_host:
+            self._ucad_host, self._ucad_port = self._ucad_host.split(":")
```
tfarago, line 271: Finally!!!! :fireworks:

concert/ext/tangoservers/reco.py
```diff
@@ -49,6 +50,14 @@ def set_{0}(self, values):
 """
 
 
+async def read_images(path):
+    with TiffSequenceReader(path) as reader:
+        image_stack = []
+        async for image in reader.read_range():
+            image_stack.append(image)
+        return image_stack
```
tfarago, line 58: This is doubled now in the local addons and here. How about putting it somewhere like `concert.readers` and using that _one_ function on both positions?

concert/helpers.py
```diff
@@ -428,3 +428,10 @@ def get_basename(filename):
         filename = filename[:-1]
 
     return os.path.basename(filename)
+
+def convert_image(image, mirror: bool, rotate: int):
+    if mirror:
+        image = np.fliplr(image)
+    if rotate:
+        image = np.rot90(image, k=rotate)
+    return image
```
tfarago, line 437: could be something like
```python
def convert_image(image, how):
    ops = {"fliplr": np.fliplr, "rot": np.rot90} # and so on
    return ops[how](image)
```

tfarago: or we can directly use the numpy function
```python
def convert_image(image, numpy_func):
    return numpy_func(image)
```

concert/networking/base.py
```diff
@@ -140,6 +142,15 @@ async def zmq_receive_image(socket):
...
     array = np.frombuffer(msg, dtype=metadata['dtype']).reshape(metadata['shape'])
 
+    mirror = False
+    rotate = 0
+    if 'mirror' in metadata:
+        mirror = metadata['mirror']
+    if 'rotate' in metadata:
+        rotate = metadata['rotate']
+
+    array = convert_image(array, mirror, rotate)
```
tfarago, line 152: and this can be
```python
getattr(numpy, metadata["conversion"])(array)
```

concert/devices/cameras/base.py
MarcusZuber: Mirror true/false + 0-3 rotations are all combinations. I have to add all 8 combinations then as choices.

concert/devices/cameras/base.py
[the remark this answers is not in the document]
MarcusZuber: You added this to the remote net camera with this interface.

bin/concert
MarcusZuber: You wanted to keep it, so one needs to think if something breaks.
(I hated it while I was debugging sessions)

concert/base.py
MarcusZuber: I found a few classes in ankaconcert that needed some fixes (because they were not done correctly). 
Otherwise, I use this since December without any other issues.

concert/devices/cameras/base.py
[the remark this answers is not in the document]
tfarago: aha, but I like that @sarkarchandan closes the sender explicitly.

bin/concert
tfarago: That's why I asked. Hm, I don't know. I mean technically if there is an error the session is not even started, so the lock file does not make sense. I will try to add the removal, OK?

concert/base.py
tfarago: OK, that pretty much convinces me.

concert/devices/cameras/base.py
tfarago: Right. I mean if  you promise me that you and no one else will want to keep adding weird transformations... :smiley:

concert/ext/tangoservers/reco.py
MarcusZuber: This is part of the commit that I would drop for now and come up with a more elegant solution.

tfarago:
Hit it.

PR #543, opened 2024-12-16
feat:  Integration testing method

sarkarchandan:
**Premise**: This PR is created to establish the foundation of a basic integration testing method for decentralized concert using GitHub Actions. It aims to automate the utility of running a concert session followed by a set of assertions about the expected outcome.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,60 @@
+FROM ubuntu:22.04
+
+ARG DEBIAN_FRONTEND=noninteractive
+
+ENV PIP_ROOT_USER_ACTION=ignore
+ENV G_MESSAGES_DEBUG=all
+ENV LD_LIBRARY_PATH=/usr/local/lib
+
+RUN apt-get update && apt-get install -y \
+    build-essential \
+    python3 \
+    python3-pip \
+    python3-setuptools \
+    python3-wheel \
+    git \
+    vim \
+    cmake \
+    libglib2.0-0 \
+    libgtk+2.0-dev \
+    gobject-introspection \
+    ninja-build \
+    libzmq5-dev \
+    libjson-glib-dev
+
+RUN pip install --upgrade pip && pip install \
+    numpy \
+    pytango==9.5.0 \
+    scikit-image \
+    meson
+
+WORKDIR /home
+
+RUN git clone https://github.com/ufo-kit/libuca.git && \
+    # For uca-net repository it is temporarily required to checkout to uca_net_grab_send branch
+    # until the same is merged to base branch.
+    git clone -b uca_net_camera_grab_send https://github.com/ufo-kit/uca-net.git
+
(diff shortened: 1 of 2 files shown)


sarkarchandan:
Python(>=3.10) tests are failing due to breaking change in scipy api.
```python
AttributeError: module 'scipy.misc' has no attribute 'ascent'
```

MarcusZuber (changes requested):
I would like to have also (I think it is called 'matrix') different OS and python versions in the tests. The rest I like! Are we sure, that the asyncio-Eventloop inside the AsyncTest is the same as in the concert-asyncio-ipython one? To be 100% sure, we could do a system call "concert start test_experiment" with a session that instantiates everithing, runs the experiment and shuts down. Then in the test, we check for valid frames.

MarcusZuber:
> > Python(>=3.10) tests are failing due to breaking change in scipy api.
> > ```python
> > AttributeError: module 'scipy.misc' has no attribute 'ascent'
> > ```
> 
> Yes I have the same issue. Just check the version and make the import correctly for old & new.

In the decentralized branch this is fixed.

Dockerfile
```diff
@@ -0,0 +1,80 @@
+FROM ubuntu:22.04
```
MarcusZuber, line 1: If we run it under ubuntu24.04 I can live without the different versions for now (on most of the beamlines I have ubuntu 24.04, or they will be upgraded soon).

concert/tests/integration/scenarios/test_experiment.py
```diff
@@ -0,0 +1,142 @@
...
+        walker_port = os.environ["REMOTE_WALKER_PORT"]
+        walker_dn = os.environ["REMOTE_WALKER_DN"]
+        self._walker_dev_uri = f"{walker_host}:{walker_port}/{walker_dn}#dbase=no"
+        self._num_darks = 10
+        self._num_flats = 10
+        self._num_radios = 100
+        self._servers = { "walker": CommData(os.environ["UCA_NET_HOST"],
+                                             port=os.environ["UCA_WALKER_ZMQ_PORT"],
+                                             socket_type=zmq.PUSH)}
+        self._camera = await RemoteNetCamera()
```
MarcusZuber, line 60: For now, we can leave everything here as it is. We can think in the future if we want to make as system call "concert start test_experiment" here, but the whole PR only improves the tests, and we do not have to overdo it.

MarcusZuber (changes requested):
Only tiny request: Please run it under ubuntu 24.04 (see comments).

I would like to add (in the future, see #544) tests with a read tango-db, but this we do later, and I think I have almost everything ready to add it without much effort.

Dockerfile
sarkarchandan: Done.

PR #545, opened 2025-03-17
Version tests

MarcusZuber:
I only wanted to test the newer python versions


MarcusZuber:
3.13 works, 3.14 still in alpha

PR #548, opened 2025-04-10
Adds qt-like signals to Parameters

MarcusZuber:
I was looking for an elegant way to connect different Parameterizable-Parameters and had this idea.

This uses https://github.com/nexconnectio/pynnex which implements qt-like signals and slots and is quite light-weight. Currently, the development is quite active and looks clean.

Example:
```python
from concert.devices.cameras.dummy import Camera
camera = await Camera()

def print_value(val):
   print(value)

camera['exposure_time'].value_set.connect(print_value)
camera.exposure = 1*q.s


exp = await CT(...)
reco = await OnlineReco(...)

exp['num_projections'].value_set.connect(reco, reco.set_number)

```

One motivation for this would be to couple experiment parameters with addon-parameters (the number of the online-reco couples with the num_projections of the CT experiment).

TODO:
- [ ] Tests
- [ ] Doc

### DIFF
--- a/concert/base.py
+++ b/concert/base.py
@@ -8,6 +8,9 @@ import functools
 import inspect
 import types
 from typing import List, Union
+
+from pynnex import emitter, with_emitters
+
 from concert.helpers import memoize, get_state_from_awaitable
 from concert.coroutines.base import background, get_event_loop, run_in_loop, wait_until
 from concert.quantities import q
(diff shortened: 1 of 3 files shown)


MarcusZuber:
Works only with python3.10+. 3.9 will be EOL this October. I don't think we will do a new realease before...

MarcusZuber:
I did' t add psutil here... Sure, lets discuss it on Friday.

PR #549, opened 2025-04-10
Require Python >= 3.8

tfarago:


MarcusZuber:
3.8 is also EOL, so we could also remove it.

tfarago:
> 3.8 is also EOL, so we could also remove it.

Nope, we still have it at ID19

PR #552, opened 2025-05-06
This allows defining session independent locks with the DEVICE_LOCKS …

MarcusZuber:
…(list of strings) variable

In a session file you can add e.g. DEVICE_LOCKS = ['my_camera1', 'reco_1_server'] to prevent the usage of devices that should be accessed only once and adding the same entry in another session file.

This exends the MULTIINSTANCE implementation. I can now specify sets of locks in session files.
The use case is, that we often use the same devices (which are somethimes not happy to be accessed multiple times) from diffrent sessions.
I can now do stuff like this:

```python
# session 1 that uses a specific camera and a remote-online-reco
DEVICE_LOCKS = ['camera1', 'reco_server1']

# session 2 that uses 'camera1' that does not like to be shared
DEVICE_LOCKS=['camera1']
```

TODO:
- [x] Add documentation
- [x] Add tests (also for the MULTIINSTANCE implementation)


MarcusZuber:
@tfarago : Do you have an idea how we could test the session-locks elegantly without system calls?

tfarago:
Looks useful.

tfarago:
> @tfarago : Do you have an idea how we could test the session-locks elegantly without system calls?

Write a mockup `StartCommand` or make a mockup of IPython.

MarcusZuber:
Some explanation about https://github.com/ufo-kit/concert/pull/552/commits/e983ad011975b8bf86ea646d08edae33d6e43770 :

`concert._aimoirt.register()` could only be called once before this commit (the paths were stored in a global variable that is not reset between tests and it does nothing if there are aready patht sepcified). The testing of the session-start triggered this the first time in the pytest run and then the `register()` in the aimport-test did nothing.

PR #554, opened 2025-05-13
Same plugin-manager is used in all uca-camera instances

MarcusZuber:


MarcusZuber:
When you ask why I didn't create the instance in the class body: Then I have to import gi.xxx globally, which we anted to prevent since we do not want to have it as hard dependency.

PR #555, opened 2025-05-14
Add experiment.remove_device_from_log

MarcusZuber:
Adds high-level function to remove devices from the logging within the experiment.


tfarago (changes requested):
Just fix the pep8 issues please and merge.

PR #567, opened 2025-10-31
feature: Alignment

sarkarchandan:
**Premise**: In this PR we implement the general alignment strategy for tomography and if possible extend for laminography.


tfarago:
> The dataclasses that with the devices and settings are elegant, but limit the amount of people at the beamline to understand whats going on a lot...

In the end there should be some higher-level functions, even a GUI, so these things are hidden completely. I will look at this Chandan when the DAQ PRs are ready.

PR #572, opened 2026-03-24
Allow limits setting/getting in parameterizable

MarcusZuber:
Since we have more and more devices where we want to store limits in the devices or in an external database, I added this to write setter/getter in the Parameterizable directly.

Like with the _set_PARAMETER it is now possible to add functions _get_PARAMETER_external_limit etc. to Parameterizables.

I also changed what the info_table prints. It now also shows target values and limits, which is currently quite cumbersome to read.

### DIFF
--- a/concert/base.py
+++ b/concert/base.py
@@ -12,7 +12,6 @@ from concert.helpers import memoize, get_state_from_awaitable
 from concert.coroutines.base import background, get_event_loop, run_in_loop, wait_until
 from concert.quantities import q
 
-
 LOG = logging.getLogger(__name__)
 _RUN_IN_LOOP_ERR_TMPL = "Someone is trying to use `{}' " \
                         "inside `async def' code which is not allowed, please " \
@@ -33,6 +32,10 @@ def _getter_not_implemented(*args):
 
 def _getter_target_not_implemented(*args):
     raise AccessorNotImplementedError
+
+
+async def _none():
+    return None
 
 
 async def _execute_func(func, instance, *args, **kwargs):
@@ -66,6 +69,7 @@ def _find_object_by_name(instance):
     """Find variable name by *instance*. This is supposed to be used only in the
     :meth:`.Parameter.__set__`.
     """
+
     def _is_name_ok(instance_name):
         return instance_name and instance_name not in ['instance', 'self']
 
@@ -131,7 +135,6 @@ class TransitionNotAllowed(FSMError):
 
 
 class StateError(Exception):
-
     """Raised in state check functions of devices."""
 
     def __init__(self, error_state, msg=None):
@@ -143,25 +146,21 @@ class StateError(Exception):
 
 
 class UnitError(ValueError):
-
     """Raised when an operation is passed value with an incompatible unit."""
     pass
 
 
 class LimitError(ValueError):
-
     """Raised when an operation is passed a value that exceeds a limit."""
     pass
 
 
 class SoftLimitError(LimitError):
-
     """Raised when a soft limit is hit on the device."""
     pass
 
 
 class HardLimitError(StateError):
-
     """Raised when device goes into hardlimit error state."""
 
     def __init__(self, error_state='hard-limit', msg=None):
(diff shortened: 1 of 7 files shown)


MarcusZuber:
Tests, doc and examples will come soon.

concert/base.py
```diff
@@ -550,6 +538,116 @@ def __init__(self, unit, fget=None, fset=None, fget_target=None, lower=None, upp
...
+    def upper_external_getter_name(self):
+        if self.external_upper_getter:
+            return self.external_upper_getter.__name__
+
+        return '_get_' + self.name + "_upper_external_limit"
+
+
+    @memoize
+    def get_lower_user_getter(self, instance):
+        if self.user_lower_getter:
```
tfarago, line 560: This is not needed, right? Handled on L550.

concert/base.py
```diff
@@ -1260,15 +1376,48 @@ def __contains__(self, key):
     async def info_table(self):
         from concert.session.utils import get_default_table
 
-        table = get_default_table(["Parameter", "Value"])
+        table = get_default_table(
+            ["Parameter", "Value", "Target", "lower limit", "upper limit", "lower user limit", "upper user limit",
+             "lower external limit", "upper external limit"])
```
tfarago: I don't like much that the table does not fit the width of my terminal unless it's full screen, is there a more compact way?

tfarago:
I would love the docs, I am starting to get lost in this. Also, I think the case when I pass the getters and setters to quantities in constructors is not set.

MarcusZuber:
> I would love the docs, I am starting to get lost in this. Also, I think the case when I pass the getters and setters to quantities in constructors is not set.

The setter/getter passing in the quantity construtor should be covered by the old tests.

concert/base.py
MarcusZuber: It is needed. The arguments of the calls are different. But the if in the _name-functions made no sense (also in the old code)

concert/base.py
MarcusZuber: When I remove the "limit" in the text it fits in a half-fullscreen terminal on my laptop:
```python
<concert.devices.motors.dummy.RotationMotor object at 0x7c3914ea6600>
 Parameter        Value                    Target  lower  upper     lower user  upper user  lower external  upper external 
 motion_velocity  50000.0 degree / second          None   None      None        None        None            None           
 position         0 degree                         None   2 degree  None        2 degree    None            None           
 state            standby         
```

MarcusZuber:
> I would love the docs, I am starting to get lost in this. Also, I think the case when I pass the getters and setters to quantities in constructors is not set.

Where do you think would be the most suited location to document this? In the Parameterizable doc?

tfarago:
> > I would love the docs, I am starting to get lost in this. Also, I think the case when I pass the getters and setters to quantities in constructors is not set.
> 
> Where do you think would be the most suited location to document this? In the Parameterizable doc?

Probably, when it comes to API. But I would definitely include it in the "user" documentation [here](https://concert.readthedocs.io/en/latest/user/topics/control.html#limits)

tfarago:
I would turn
```python
    @memoize
    def upper_external_getter_name(self):
        return '_get_' + self.name + "_upper_external_limit"
```

into

```python
    @property
    def upper_external_getter_name(self):
        return '_get_' + self.name + "_upper_external_limit"
```

It is a typical case of a calculated property.

MarcusZuber:
> > > I would love the docs, I am starting to get lost in this. Also, I think the case when I pass the getters and setters to quantities in constructors is not set.
> > 
> > 
> > Where do you think would be the most suited location to document this? In the Parameterizable doc?
> 
> Probably, when it comes to API. But I would definitely include it in the "user" documentation [here](https://concert.readthedocs.io/en/latest/user/topics/control.html#limits)

I will add a short description what the idea behind the different limits is and add a link to the API examples.

MarcusZuber:
Here is the corresponding new documentation:
https://concert.readthedocs.io/en/add-limit-handling-in-parameterizable/

tfarago:
:+1:

PR #573, opened 2026-04-14
Make reco args in tango-server instance variable

MarcusZuber:
The reco args in the reco tango server were a class attribute which messed up the attributes when having multiple reco devices in the same device server instance. Changing parameters in one changed it for all.


MarcusZuber:
At some point I will remove the `exec` (which I dislike since the beginning), but for now this works now with multiple devices.

tfarago:
:+1:

tfarago (approved):

PR #577, opened 2026-05-31
Implement projection based FRC

sarkarchandan:
**Premise**: Implement projection based FRC


tfarago:
There should be a possibility to just compute FRC between two images without the need of specifying dark and flat.

